PR #184, opened 2026-05-23
Fix/back to top event cleanup 130

### LINKED ISSUE #130: [BUG]: Memory Leak in App.jsx — Missing click event listener cleanup on Back-to-Top button
### Bug Summary
While reviewing the `App.jsx` component, I discovered a classic React memory leak in the `useEffect` hook that handles the "Back to Top" button logic. 

The code successfully attaches two event listeners (`scroll` on the window object, and `click` on the button DOM element). However, the cleanup function returned by the hook only removes the `scroll` listener, completely failing to remove the `click` listener. 

### Current Behavior
Whenever the `App` component unmounts and remounts, a new `click` listener is added to the DOM node without the previous one being destroyed. This creates duplicated listeners firing simultaneously, leading to memory bloat and potential performance degradation.

### Expected Behavior
Every event listener attached manually to the DOM inside a `useEffect` must be properly cleaned up using `removeEventListener` when the component unmounts.

### Steps to Reproduce
1. Navigate to `src/App.jsx` around line 220.
2. Observe the following code block:
```javascript
  useEffect(()=>{
    const btn = document.getElementById('back-to-top');
    if (!btn) return;
    const fn = () => btn.classList.toggle('visible', window.scrollY > 400);
    window.addEventListener('scroll', fn, { passive:true });
    
    // BUG: This listener is added...
    btn.addEventListener('click', () => window.scrollTo({ top:0, behavior:'smooth' }));
    
    // ...but only the scroll listener is removed here!
    return () => window.removeEventListener('scroll', fn); 
  }, []);

anshika1179:
Closes #130
## Summary
* fixed memory leak in Back-to-Top button logic
* added proper cleanup for click event listeners
* improved useEffect lifecycle management
* prevented duplicated event handler registration

## Technical Changes
* replaced anonymous click listener with named handler
* cleaned both scroll and click listeners correctly
* improved defensive DOM handling
* preserved smooth scroll behavior

## Performance Improvements
* prevents duplicated event listeners
* reduces memory growth during remounts
* improves long-session frontend stability
* maintains efficient scroll handling

## Verification
* verified event listeners are cleaned correctly
* confirmed smooth scroll behavior still works
* tested repeated mount/unmount cycles
* validated no duplicate click handlers remain


Ayushh-Sharmaa:

server-python/requirements.txt

PR #234, opened 2026-05-24
fix: prevent portfolio DB fallback split-brain and fail-closed auth

### LINKED ISSUE #230: Portfolio DB fallback “fails open” + permanent split-brain (unauthorized overwrites during DB outages)
## Problem

`server/repositories/portfolioRepository.js` implements a DB-first storage layer with a local JSON file fallback. The current fallback design has two compounding problems:

### 1) DB availability is cached “forever” after the first failure

`ensureReady()` stores a `schemaReady` promise and, on the first DB error, permanently sets:

- `schemaReady = Promise.resolve(false)`

That means a transient DB outage at startup (or the first time `ensureReady()` runs) can flip the process into local-file mode for the rest of its lifetime, even after the DB becomes healthy again. This creates a long-lived split-brain between:
- the DB (source of truth), and
- `server/data/portfolios.json` (fallback store)

### 2) Auth decision can “fail open” on fallback

`verifyPasskey(username, passkey)` returns `true` when a username does not exist:

- DB path: `if (!rows.length) return true; // new registration`
- Local fallback path: `if (!portfolio) return true; // new registration`

In a split-brain scenario (DB contains a username, but local file does not), DB failures cause fallback to return `true`, treating an existing username as “new registration”.

Combined with `createOrUpdate()` behavior, this can allow an attacker to overwrite an existing portfolio during DB instability by presenting *any* passkey, because the auth check may incorrectly allow the operation.

## Why this is critical

- **Integrity loss:** legitimate portfolios can be overwritten or diverge across stores.
- **Account takeover path during outages:** authorization logic makes “new registration” decisions without a reliable global existence check.
- **Operational fragility:** one temporary DB incident can push the server into a long-lived degraded mode with silent data corruption.

## Example failure scenario

1. DB has portfolio for `alice`.
2. Local `portfolios.json` does *not* have `alice` (fresh instance, cleared file, different node, container restart, etc.).
3. DB becomes unavailable (network blip / credentials rotate / maintenance).
4. Requests now fall back to local file mode.
5. `verifyPasskey('alice', 'anything')` returns `true` because local file lacks `alice`.
6. The system proceeds as if it’s a new registration/update and writes inconsistent state.

## Expected behavior

- If DB is configured as primary, DB outages should not silently switch to an inconsistent writeable store.
- Auth should **fail closed** when the system cannot reliably determine whether an account exists and whether a credential matches.

1. Replace the “forever cached” DB availability with a health-check + TTL approach:
   - cache success/failure for a short period (e.g., 5–30s), then retry.
2. Split read and write strategies:
   - allow local fallback as **read-only** cache (optional),
   - fail writes when DB is down (or enqueue writes for later reconciliation).
3. Make `verifyPasskey()` fail closed when primary store is unavailable:
   - do **not** return `true` for “new registration” on fallback unless you can prove non-existence globally.
4. Introduce a single authoritative “exists(username)” check with strict semantics.
5. Add concurrency safety for the local file (atomic writes + locking) if it remains in use.
6. Add tests simulating:
   - transient DB failure then recovery,
   - split-brain (DB has user, local file doesn’t),
   - ensuring updates are rejected rather than silently accepted.

ionfwsrijan:
---

## Problem

`server/repositories/portfolioRepository.js` has two compounding bugs:

1. **DB failure cached forever** — `ensureReady()` permanently sets `schemaReady = Promise.resolve(false)` on the first DB error. A transient outage at startup flips the process into local-file mode for its entire lifetime, creating long-lived split-brain.

2. **`verifyPasskey()` fails open** — returns `true` ("new registration") when a username doesn't exist in the local fallback. During split-brain (DB has user, local file doesn't), this allows anyone to overwrite an existing portfolio with any passkey.

3. **`createOrUpdate()` writes to local file** when DB is down, persisting the split-brain state.

## Changes

1. **TTL-based health check** — replaced forever-cached `schemaReady` with a 15-second retry window. After a DB failure, `ensureReady()` re-tests with a lightweight `SELECT 1` every 15s. Transient outages no longer permanently disable the DB.

2. **`verifyPasskey()` fails closed** — local file fallback now returns `false` (deny) when the user is not found in the local cache, instead of `true` (allow). The caller responds with 401, preventing silent overwrites.

3. **`createOrUpdate()` rejects writes when DB is down** — local file write fallback is removed entirely. Throws `"Portfolio storage is unavailable. Please try again later."` instead of silently persisting to an inconsistent store.

4. **Removed unused `writeLocalPortfolios()`** — no longer needed.

`getByUsername()` still uses the local file as a read-only cache during DB outages (stale reads are safe).

Fixes #230


ionfwsrijan:
@Ayushh-Sharmaa You may review and merge

Ayushh-Sharmaa:

server/repositories/portfolioRepository.js

Ayushh-Sharmaa:
@ionfwsrijan Any Update

ionfwsrijan:
@Ayushh-Sharmaa My bad i forgot about this. Give me just 5mins. In the meanwhile you can merge my other PRs

ionfwsrijan:
@Ayushh-Sharmaa The conflicts have been resolved. You may merge this now

PR #235, opened 2026-05-24
fix: introduce PUBLIC_APP_URL env var to fix broken verify links with multi-origin CORS_ORIGIN

### LINKED ISSUE #231: Broken verify links when CORS_ORIGIN contains multiple origins (email verification unusable in real deploys)
## Problem

`CORS_ORIGIN` is treated as a comma-separated allowlist in CORS config:

- `server/index.js`: `process.env.CORS_ORIGIN.split(',').map(...).filter(Boolean)`

But the welcome/verification email link is built using the raw `CORS_ORIGIN` string as if it were a single base URL:

- `server/index.js`: `const verifyUrl = \`${process.env.CORS_ORIGIN || 'http://localhost:5173'}/verify?email=...\`;`

If `CORS_ORIGIN` is set to multiple origins (a common production setup), the generated `verifyUrl` becomes invalid, e.g.:

- `https://app.example.com,https://www.example.com/verify?email=...`

This breaks onboarding/verification flows and causes hard-to-debug user-facing failures.

## Why this is critical

- **Core user flow breakage:** users can’t verify emails (or land on a valid verification route).
- **Production-only bug:** it won’t reproduce in dev where `CORS_ORIGIN` is typically a single localhost URL.
- **Configuration ambiguity:** the same env var is used for two different meanings (allowlist vs canonical public base URL).

## Reproduction

1. Set `CORS_ORIGIN="https://app.example.com,https://www.example.com"`.
2. Submit a membership/recruitment/core-team form (anything that triggers the welcome email in `handleForm()`).
3. Inspect the email button/link URL.
4. Observe the URL is malformed and does not resolve.

## Expected behavior

Verification URLs should be based on a single canonical public base URL, not an allowlist.

1. Introduce a dedicated env var, e.g. `PUBLIC_APP_URL` (single URL).
2. Validate it at startup:
   - must be a valid absolute URL,
   - must not contain commas,
   - optional: must match one of the allowed CORS origins.
3. Use `PUBLIC_APP_URL` to construct:
   - welcome verification links,
   - password reset links (if applicable),
   - SSE allowed origin logic (avoid “first origin wins” behavior in `server/services/sseService.js`).
4. Add config docs and an `.env.example` update.
5. Add tests for `handleForm()` URL building under multi-origin `CORS_ORIGIN`.

ionfwsrijan:
## Problem

`CORS_ORIGIN` is treated as a comma-separated allowlist in the CORS config (`process.env.CORS_ORIGIN.split(',')`), but the welcome/verification email link uses the raw `CORS_ORIGIN` string as if it were a single base URL:

```js
const verifyUrl = `${process.env.CORS_ORIGIN || 'http://localhost:5173'}/verify?email=...`;
```

If `CORS_ORIGIN` is set to multiple origins (common in production), the generated `verifyUrl` becomes malformed — e.g. `https://app.example.com,https://www.example.com/verify?email=...` — which breaks email verification entirely.

The same "first origin wins" anti-pattern exists in the SSE service's `setupSSEHeaders()`, which picks only the first origin from the comma-separated list.

## Changes

1. **New `PUBLIC_APP_URL` env var** — a single canonical base URL for building app links. Validated at startup: rejects commas and invalid absolute URLs.
2. **`getPublicAppUrl()` helper** — resolves `PUBLIC_APP_URL` → first `CORS_ORIGIN` → `http://localhost:5173` fallback.
3. **`server/index.js`** — verify URL in `handleForm()` now uses `getPublicAppUrl()` instead of raw `CORS_ORIGIN`.
4. **`server/services/sseService.js`** — SSE `Access-Control-Allow-Origin` prefers `PUBLIC_APP_URL` before falling back to the first CORS origin.
5. **`server/.env.example`** — documented `PUBLIC_APP_URL`.

Backward compatible: if `PUBLIC_APP_URL` is not set, behavior falls back to the first `CORS_ORIGIN` entry or `http://localhost:5173`.

Fixes #231

### DIFF
--- a/server/.env.example
+++ b/server/.env.example
@@ -4,6 +4,8 @@
 PORT=8787
 NODE_ENV=development
 CORS_ORIGIN=http://localhost:5173
+# Canonical public URL for building email verification and other app links (single URL, not comma-separated). Falls back to the first CORS_ORIGIN if not set.
+PUBLIC_APP_URL=http://localhost:5173
 
 # Database & Authentication (Supabase)
 SUPABASE_URL=
--- a/server/index.js
+++ b/server/index.js
@@ -109,6 +109,25 @@ function requiredStrongPassword(name) {
 
 const ADMIN_EVENT_PASSWORD = requiredStrongPassword('ADMIN_EVENT_PASSWORD');
 
+if (process.env.PUBLIC_APP_URL) {
+  if (process.env.PUBLIC_APP_URL.includes(',')) {
+    throw new Error('PUBLIC_APP_URL must be a single URL, not a comma-separated list');
+  }
+  try {
+    new URL(process.env.PUBLIC_APP_URL);
+  } catch {
+    throw new Error(`PUBLIC_APP_URL must be a valid absolute URL, got: ${process.env.PUBLIC_APP_URL}`);
+  }
+}
+
+function getPublicAppUrl() {
+  if (process.env.PUBLIC_APP_URL) {
+    return process.env.PUBLIC_APP_URL.replace(/\/+$/, '');
+  }
+  const firstOrigin = process.env.CORS_ORIGIN?.split(',')[0]?.trim();
+  return firstOrigin || 'http://localhost:5173';
+}
+
 function normalizePrivateKey(k) {
   return k.includes('\\n') ? k.replace(/\\n/g, '\n') : k;
 }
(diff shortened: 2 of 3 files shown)


ionfwsrijan:
@Ayushh-Sharmaa You may review and merge this

PR #236, opened 2026-05-24
fix: remove token query param auth support from admin endpoints

### LINKED ISSUE #229: Admin auth accepts ?token= query param (token leakage + bypass-prone pattern)
## Problem

`requireAdmin` accepts an admin session token from a URL query parameter:

- `server/middleware/adminAuthMiddleware.js`: `parseBearer(...) || req.query.token`

This means any `/api/admin/*` route protected by `requireAdmin` can be called as:

- `/api/admin/...?...&token=<ADMIN_SESSION_TOKEN>`

## Why this is critical

Putting bearer-like secrets in URLs is broadly unsafe because URLs are commonly:
- logged by reverse proxies/CDNs/WAFs and server access logs,
- stored in browser history,
- leaked via `Referer` headers when the page makes requests to other origins,
- shared unintentionally via screenshots, copy/paste, chat, etc.

Even if the frontend *doesn't* currently use `?token=`, the backend support encourages insecure usage and increases the blast radius of any accidental token exposure.

## Affected surface

Anything behind `adminAuthMiddleware.requireAdmin`, including (non-exhaustive):
- `/api/admin/metrics/*`
- `/api/admin/events/*`
- `/api/admin/core-team/*`
- `/api/admin/analytics/*`

(See `server/index.js` where `adminAuth` is applied.)

## Reproduction (local)

1. Log in as admin via `POST /api/admin/login` and get a token in the JSON response.
2. Call a protected admin endpoint using a URL token:
   - `GET /api/admin/events?token=<token>`
3. Observe it succeeds without an `Authorization: Bearer <token>` header.

## Expected behavior

Admin session tokens should be accepted only via:
- `Authorization: Bearer <token>` header (minimum), or
- a secure `HttpOnly` cookie-based session (if/when UI is added).

1. Remove query-param token support entirely:
   - delete `|| req.query.token` in `requireAdmin`.
2. Add a hard fail-fast guard:
   - if `req.query.token` exists, return `400` with a message like “Do not pass tokens in URLs”.
3. Update any clients/docs/swagger that might rely on URL tokens.
4. Add tests:
   - unit test for middleware rejecting `req.query.token`,
   - integration test for `/api/admin/*` routes requiring `Authorization` header.
5. (Optional hardening) Add request logging redaction for `Authorization` header and any query keys named `token` to prevent accidental leakage elsewhere.

ionfwsrijan:
## Problem

`requireAdmin` middleware accepts an admin session token from a URL query parameter (`req.query.token`), which is a security risk because:

- URLs are logged by reverse proxies, CDNs, WAFs, and server access logs
- URLs are stored in browser history
- Tokens can leak via `Referer` headers when the page makes requests to other origins
- URLs can be shared unintentionally via screenshots, copy/paste, or chat

Even if the frontend primarily uses `Authorization: Bearer` headers, the backend support for `?token=` encourages insecure usage and increases the blast radius of any accidental token exposure.

## Changes

1. **`server/middleware/adminAuthMiddleware.js`** — Removed `|| req.query.token` fallback from `requireAdmin()`. Added a fail-fast guard: if `req.query.token` is present, returns `400` with the message `"Do not pass tokens in URLs. Use Authorization: Bearer header."`

2. **`src/pages/admin/AdminPage.jsx`** — The admin dashboard used `EventSource` (SSE) with `?token=` in the URL because the browser API doesn't support custom headers. Replaced it with a fetch-based SSE client that sends `Authorization: Bearer` via the standard `headers` option, while preserving auto-reconnect behavior on connection drops.

3. **`server/index.js`** — Added `redactUrl()` to the request logger to redact `token` query parameter values from logged URLs, preventing accidental token leakage in server/access logs.

Fixes #229

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -31,6 +31,10 @@ app.use(cors({
 }));
 app.use(express.json({ limit: '512kb' }));
 
+function redactUrl(url) {
+  return url.replace(/[?&]token=[^&\s]+/gi, '$1token=REDACTED');
+}
+
 function requestLogger(req, res, next) {
   const start = process.hrtime.bigint();
   const { method, path } = req;
@@ -38,7 +42,9 @@ function requestLogger(req, res, next) {
   res.on('finish', () => {
     const duration = Number(process.hrtime.bigint() - start) / 1e6;
     const status = res.statusCode;
-    const message = `[${method}] ${path} → ${status} (${Math.round(duration)}ms)`;
+    const redactedPath = redactUrl(path);
+    const redactedUrl = req.originalUrl ? redactUrl(req.originalUrl) : redactedPath;
+    const message = `[${method}] ${redactedUrl} → ${status} (${Math.round(duration)}ms)`;
 
     if (status >= 500) {
       console.error(message);
--- a/server/middleware/adminAuthMiddleware.js
+++ b/server/middleware/adminAuthMiddleware.js
@@ -80,7 +80,11 @@ function parseBearer(authHeader = '') {
 
 async function requireAdmin(req, res, next) {
   try {
-    const bearer = parseBearer(req.headers.authorization || '') || req.query.token;
+    if (req.query.token) {
+      return res.status(400).json({ error: 'Do not pass tokens in URLs. Use Authorization: Bearer header.' });
+    }
+
+    const bearer = parseBearer(req.headers.authorization || '');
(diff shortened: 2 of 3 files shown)


ionfwsrijan:
@Ayushh-Sharmaa You may review and merge.

PR #241, opened 2026-05-24
fix: strip creator PII from public activity event API

### LINKED ISSUE #179: Activity event creator PII (name, email, phone) leaked through public API when using file-based storage
## Problem Description
When Supabase is not configured (or Supabase initialization fails), the system falls back to file-based storage (`content.json`). The file-based storage path for activity events stores and returns the event creator's full PII — including name, email, and phone number — through the fully public `GET /api/content/activity-events/:activityKey` endpoint.

### The Creation Path (stores PII)
```js
// server/index.js:521-553 — POST /api/content/activity-events/:activityKey
const event = {
  // ...
  createdBy: {
    name: toSafeString(body.name, 120),
    email: toSafeString(body.email, 140),
    phone: normalizePhone(body.phone),
  },
};
```

### The Retrieval Path (returns PII — local file path)
```js
// server/index.js:302-317
async function listActivityEventsStore(activityKey) {
  // ...
  // For local file storage:
  const content = await readContent();
  return (content.activityEvents?.[activityKey] || []).map((event) => sanitizeActivityEventRecord(event));
}

// server/index.js:319-321 — sanitize function does NOTHING:
function sanitizeActivityEventRecord(event) {
  return event;
}
```

### The Supabase Path (correctly strips PII)
For comparison, the Supabase query at line 303-313 intentionally omits the `created_by_` fields:
```js
return rows.map(r => sanitizeActivityEventRecord({
  id: r.id, name: r.name, date: r.date_text || r.date,
  tagline: r.tagline, description: r.description,
  status: r.status || 'completed', createdAt: r.created_at,
  // NOTE: created_by_name, created_by_email, created_by_phone are NOT included
}));
```

## Technical Root Cause
1. The event creation handler (`POST /api/content/activity-events/:activityKey`) always stores `createdBy` with name, email, and phone
2. The `sanitizeActivityEventRecord` function at line 319 is a no-op identity function — it returns its input unchanged
3. The Supabase code path correctly strips PII by not including the `created_by_` fields in the mapped output, but the local file path returns the full stored object including `createdBy`
4. This means the security of PII depends on which storage backend is active, creating an invisible vulnerability

## Affected Components
- `server/index.js:319-321` — `sanitizeActivityEventRecord()` (no-op, should be stripping `createdBy`)
- `server/index.js:302-317` — `listActivityEventsStore()` (local file path)
- `server/index.js:521-553` — `POST /api/content/activity-events/:activityKey` (PII storage)
- `server/index.js:512-519` — `GET /api/content/activity-events/:activityKey` (public PII exposure)

## Reproduction Scenario
1. Deploy without Supabase configured (falls back to `content.json`)
2. A core team member creates an activity event via `POST /api/content/activity-events/my-activity` with their name, email, and phone
3. Anyone calls `GET /api/content/activity-events/my-activity`
4. Response includes `createdBy: { name: "...", email: "...", phone: "..." }`

## Expected Behavior
The public API should never return PII of event creators. The `sanitizeActivityEventRecord` function should strip `createdBy` (or at minimum remove email and phone) before returning data through a public endpoint.

## Actual Behavior
Full PII of activity event creators is returned through a public, unauthenticated endpoint when using file-based storage. The same endpoint when using Supabase works correctly because the Supabase code path simply omits the fields.

## Production Impact
- Core team member email addresses and phone numbers are publicly exposed
- This enables targeted phishing, spam, and social engineering attacks against event organizers
- Unlike Issue #114 (which was about core team data and has been addressed), this exposes PII through a different endpoint and data type (activity events)
- The vulnerability is invisible to developers testing with Supabase (where it works correctly) and only manifests in file-backed deployments

1. Make `sanitizeActivityEventRecord` actually sanitize: strip the `createdBy` field (or at minimum email and phone) from public API responses
2. Alternatively, split the storage to keep PII separate from public event data
3. Add a Supabase-only conditional check similar to how other store functions handle it
4. Consider whether the `createdBy` information is even necessary for the public endpoint — the existing auth-based flow already verifies the creator during event creation

atul-upadhyay-7:
Closes #179

## Root Cause
sanitizeActivityEventRecord in server/index.js was a no-op — it returned the full event including createdBy with name, email, phone. The Supabase code path worked correctly (intentionally omitted created_by_ fields), but the file-based storage path exposed PII through GET /api/content/activity-events/:key.

## Fix
sanitizeActivityEventRecord now strips the createdBy field before returning. Applied to both server/index.js (the function used at runtime) and server/utils/sanitize.js (tested utility).

## Verification
- PII still stored in content.json for audit
- PII stripped from all public API responses
- All 4 existing tests pass

## Suggested Labels
security, bug

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -360,7 +360,9 @@ async function listActivityEventsStore(activityKey) {
 }
 
 function sanitizeActivityEventRecord(event) {
-  return event;
+  if (!event || typeof event !== 'object') return event;
+  const { createdBy, ...safe } = event;
+  return safe;
 }
 
 async function createActivityEventStore(activityKey, event) {
--- a/server/test/sanitize.test.js
+++ b/server/test/sanitize.test.js
@@ -55,7 +55,7 @@ test('sanitizeCoreTeamMemberRecord escapes profile fields and optional links', (
   assert.equal(sanitized.photoUrl, 'https://cdn.example.com/&quot;avatar&quot;.png?alt=&lt;img&gt;');
 });
 
-test('sanitizeActivityEventRecord escapes nested creator fields', () => {
+test('sanitizeActivityEventRecord strips createdBy PII', () => {
   const sanitized = sanitizeActivityEventRecord({
     name: '<script>Session</script>',
     date: 'June 1, 2026',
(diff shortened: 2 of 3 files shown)


atul-upadhyay-7:
@Ayushh-Sharmaa  could you merge this now

PR #321, opened 2026-05-25
fix: preserve matching CORS origin for admin SSE stream

### LINKED ISSUE #281: Bug: Admin SSE stream breaks on secondary allowed origins because Access-Control-Allow-Origin is pinned to the first CORS entry
## Summary

The main Express CORS middleware correctly accepts every origin listed in `CORS_ORIGIN`, but the SSE middleware does not reuse that logic. `setupSSEHeaders()` always picks only the first origin from the comma-separated list and hardcodes it into `Access-Control-Allow-Origin`. As a result, the admin dashboard can log in from a secondary allowed origin, but its live `EventSource` stream still fails CORS.

## Affected files

- `server/index.js` (global CORS allowlist at lines 29-31)
- `server/services/sseService.js` (`setupSSEHeaders`, lines 61-70)
- `src/pages/admin/AdminPage.jsx` (SSE connection at lines 79-81)

## Reproduction

1. Set `CORS_ORIGIN` to multiple values, for example:
   - `https://prod.example.com,https://preview.example.com`
2. Open the admin dashboard from `https://preview.example.com`.
3. Log in successfully. Regular authenticated fetches can work because the main `cors()` middleware accepts both configured origins.
4. Let the dashboard open its live stream connection:
   - `GET /api/admin/metrics/stream?token=...`
5. Inspect the SSE response headers. `setupSSEHeaders()` emits:
   - `Access-Control-Allow-Origin: https://prod.example.com`
6. The browser rejects the EventSource connection on `https://preview.example.com`, so live registration/login updates never arrive.

## Expected behavior

The SSE response should allow whichever request origin is actually making the call, as long as that origin is present in the configured allowlist.

## Actual behavior

Only the first configured origin can consume the admin SSE stream. Secondary preview/staging origins fail even though they are otherwise allowed by the app-level CORS configuration.

## Why this is an advanced issue

This is a transport-specific configuration drift bug that only appears in multi-origin deployments. It is easy to miss locally, but it breaks real-time admin features in preview or secondary production environments.

## Acceptance criteria

- SSE origin validation reuses the same multi-origin allowlist semantics as the main Express CORS middleware.
- The response echoes the matching request origin instead of pinning to the first configured entry.
- Admin live updates work from every origin explicitly listed in `CORS_ORIGIN`.
- Add regression coverage for multi-origin SSE behavior.

shashank03-dev:
## Description

This fixes the admin SSE CORS bug when `CORS_ORIGIN` contains more than one allowed origin. The global `cors()` middleware was already picking the right origin for each request, but `setupSSEHeaders()` replaced that header with `PUBLIC_APP_URL` or the first entry from `CORS_ORIGIN`. That meant a preview or secondary origin could log in successfully and still lose the live admin stream.

The fix leaves origin selection to the existing app-level CORS middleware and adds a regression test for both a secondary allowed origin and a disallowed origin.

Verified locally with:

- `cd server && node --test test/sseService.test.js`
- `cd server && node --test test/*.test.js`

## Related Issue

Closes #281

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Chore or maintenance

## Checklist

- [x] I have tested these changes locally.
- [x] I have run the relevant lint, build, or test commands.
- [x] I have linked the related issue.
- [x] My changes follow the existing project style.

## Screenshots

Not applicable. This change only affects backend SSE response headers and test coverage.

### DIFF
--- a/server/services/sseService.js
+++ b/server/services/sseService.js
@@ -59,15 +59,12 @@ export function getConnectedSSEClientsCount() {
  * SSE middleware setup
  */
 export function setupSSEHeaders(req, res, next) {
-  const allowedOrigin = process.env.PUBLIC_APP_URL || process.env.CORS_ORIGIN?.split(',')[0]?.trim() || 'http://localhost:5173';
-
   res.setHeader('Content-Type', 'text/event-stream');
   res.setHeader('Cache-Control', 'no-cache');
   res.setHeader('Connection', 'keep-alive');
-  res.setHeader('Access-Control-Allow-Origin', allowedOrigin);
-  res.setHeader('Access-Control-Allow-Methods', 'GET, OPTIONS');
-  res.setHeader('Access-Control-Allow-Headers', 'Content-Type, Authorization');
-  res.setHeader('Access-Control-Allow-Credentials', 'true');
+
+  // The app-level cors() middleware already selected the correct origin.
+  // Do not overwrite it here, or multi-origin deployments break.
 
   // Send initial connection message
   res.write(': SSE connection established\n\n');
--- a/server/test/sseService.test.js
+++ b/server/test/sseService.test.js
@@ -1,0 +1,89 @@
+import assert from 'node:assert/strict';
+import { once } from 'node:events';
+import http from 'node:http';
+import test from 'node:test';
+
+import cors from 'cors';
+import express from 'express';
+
+import { setupSSEHeaders } from '../services/sseService.js';
+
+function buildCorsOriginOption() {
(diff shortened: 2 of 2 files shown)


Ayushh-Sharmaa:

server/services/sseService.js

shashank03-dev:
@Ayushh-Sharmaa I've Resolved the conflict in `server/services/sseService.js`, synced the branch with the latest `main`, and pushed the update in commit `26fa27f`.

PR #323, opened 2026-05-25
fix: add mutex lock to portfolioRepository fallback writes

### LINKED ISSUE #272: Bug: Concurrent File-Based Portfolio Writes Can Cause Silent Data Loss
## Description

When PostgreSQL becomes unavailable, the backend falls back to JSON file storage.

Unlike `content.json` updates (which already use a mutex lock), portfolio writes inside `portfolioRepository.js` execute without synchronization.

Current flow:

```js
readLocalPortfolios()
writeLocalPortfolios()
```

Because these operations are asynchronous and unguarded, concurrent requests can overwrite one another.

---

## Impact

During database outages or fallback mode, simultaneous portfolio updates or registrations can silently lose data.

Under concurrent traffic:
- one request overwrites another
- registrations disappear
- portfolio state becomes inconsistent

This creates a serious data consistency risk during degraded backend conditions.

---

## Steps to Reproduce

1. Disable PostgreSQL to force fallback mode.

2. Trigger concurrent `createOrUpdate()` requests using `Promise.all`.

Example:

```js
await Promise.all([
  createOrUpdate(userA),
  createOrUpdate(userB)
]);
```

3. Inspect `portfolios.json`.

4. Observe that one of the entries may disappear due to write races.

---

## Expected Behavior

Concurrent file writes should execute sequentially or under a shared mutex lock.

---

## Actual Behavior

Concurrent writes race against each other and overwrite previous file state.

---

## Proposed Solution

Reuse the existing mutex strategy already used for `content.json` updates.

Possible approaches:
- extract `withContentLock` into a shared utility
- introduce a dedicated portfolio write lock
- serialize fallback storage writes
- queue file write operations during fallback mode

This ensures file-based fallback storage remains consistent under concurrency.

bhumikasudarshani-cmd:
closes #272

## Changes
- Added `Mutex` import from `async-mutex`
- Created dedicated `portfolioMutex` instance
- Wrapped local file fallback in `createOrUpdate()` with `portfolioMutex.runExclusive()`

## Reason
Portfolio writes in fallback mode were unguarded, causing race conditions and silent data loss. 
This follows the same mutex pattern already used for `content.json` writes in `index.js`.

### DIFF
--- a/server/repositories/portfolioRepository.js
+++ b/server/repositories/portfolioRepository.js
@@ -3,10 +3,12 @@ import { promises as fs } from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
 import { withDb } from './db.js';
+import { Mutex } from 'async-mutex';
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 const PORTFOLIOS_FILE = path.join(__dirname, '..', 'data', 'portfolios.json');
+const portfolioMutex = new Mutex();
 
 const BCRYPT_ROUNDS = 12;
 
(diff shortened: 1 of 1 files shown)


Ayushh-Sharmaa:

server/repositories/portfolioRepository.js

PR #383, opened 2026-05-26
Hash passwords with bcrypt and session tokens with SHA-256

### LINKED ISSUE #379: [Security] Java Backend: Plaintext Password Comparison + In-Memory Unhashed Admin Session Tokens
## Problem Description

The Java (Spring Boot) backend uses TWO critical authentication anti-patterns:

1. **Plaintext password comparison**: Admin login credentials are compared directly as raw strings with no hashing whatsoever.
2. **In-memory token storage with no hashing**: Session tokens are stored as-is in a `ConcurrentHashMap` — no hashing, no persistence, no rotation.

This means: if an attacker gains any memory access (heap dump, debug port, container escape), they instantly have all active admin tokens in plaintext AND can authenticate knowing only the raw password. The Node.js backend correctly hashes tokens via SHA-256 (`adminSessionsRepository.js:22`) and persists sessions to PostgreSQL — but the Java backend does neither.

### Issue 3a: Plaintext Password Comparison (AdminAuthService.java:34-36)

```java
@Value("${ADMIN_EMAIL}")
private String adminEmail;

@Value("${ADMIN_PASSWORD}")
private String adminPassword;

public boolean isValidCredentials(String email, String password) {
    return adminEmail.equals(email) && adminPassword.equals(password);
}
```

The injected `adminPassword` value (from `application.properties` or environment) is compared directly against the user-supplied password using `String.equals()`. There is no bcrypt, no scrypt, no Argon2 — just a raw string comparison. Anyone who can read the application configuration (via env dump, config server leak, or debug endpoint) gets the actual password.

### Issue 3b: Unhashed In-Memory Token Storage (TokenService.java:23-31)

```java
private final ConcurrentHashMap<String, SessionInfo> sessions = new ConcurrentHashMap<>();

public TokenSession createSession(String email) {
    Instant now = Instant.now();
    String token = generateToken();
    SessionInfo sessionInfo = new SessionInfo(token, email, now, now.plus(SESSION_TTL));
    sessions.put(token, sessionInfo);
    ...
}
```

1. Tokens are stored **as plaintext keys** in the HashMap — no SHA-256 hashing (compare with Node.js `adminSessionsRepository.js:22` which does `crypto.createHash('sha256').update(String(token)).digest('hex')`)
2. Sessions are **entirely in-memory** with no database persistence — every JVM restart invalidates ALL active sessions
3. The `ConcurrentHashMap` grows unboundedly until the scheduled cleanup runs (every 30 minutes) — a burst of login requests between cleanups can cause OOM
4. Tokens are 48 hex characters (192 bits of entropy), but since they're stored in plaintext, a memory dump reveals all active tokens

### Contrast with Node.js implementation

The Node.js backend (`server/repositories/adminSessionsRepository.js`) correctly:
- Hashes tokens with SHA-256 before storage (line 22)
- Persists sessions in PostgreSQL (line 26-36)
- Runs periodic cleanup on a configurable timer (line 142-153)
- Stores session creation/revocation timestamps
- Cleans up expired sessions on query (line 92-99)

The Java implementation has none of these protections.

## Affected Components/Files

- `server-java/src/main/java/org/nexasphere/service/AdminAuthService.java:34-36` — plaintext password comparison
- `server-java/src/main/java/org/nexasphere/service/TokenService.java:23-31` — plaintext in-memory token storage
- `server-java/src/main/java/org/nexasphere/config/BearerTokenAuthFilter.java:43-48` — token validation from in-memory store
- `server-java/src/main/java/org/nexasphere/controller/AdminController.java:40-41` — login endpoint
- `server-java/src/main/resources/application.properties` — where the admin password is configured
- `server-java/pom.xml` — needs bcrypt/Spring Security Crypto dependency

## Reproduction Scenario

1. Deploy the Java backend with any `ADMIN_PASSWORD` value
2. As an attacker with access to the server process:
   - Take a heap dump: `jmap -dump:live,format=b,file=heap.bin <PID>`
   - Search for all `SessionInfo` objects in the dump — every admin token is visible in plaintext
   - Use any extracted token to call `GET /api/admin/me` or `GET /api/admin/events` with `Authorization: Bearer <token>`
3. Alternatively, if the attacker gets access to the environment config (e.g., via `/actuator/env` if actuator is exposed, or a CI/CD secret leak), the actual admin password is directly usable — no hash cracking needed
4. Restart the Java backend process — all sessions are instantly invalidated, forcing every admin to re-login, but no token revocation tracking exists

## Expected Behavior

- Passwords should be hashed with bcrypt/Argon2 before any comparison, following Spring Security's standard `PasswordEncoder` pattern
- Session tokens should be SHA-256 hashed before storage, matching the Node.js backend pattern
- Admin sessions should be persisted to the database for durability across restarts
- Token revocation should be tracked and enforced

## Actual Behavior

- Plaintext password comparison: `adminPassword.equals(password)` at `AdminAuthService.java:36`
- Plaintext token storage: `sessions.put(token, sessionInfo)` at `TokenService.java:30`
- No persistence: `ConcurrentHashMap` is memory-only, lost on each restart

## Production Impact

- **Complete admin account compromise**: Any memory read (heap dump via JMX, /actuator/heapdump, container escape) reveals ALL active session tokens AND the admin password in plaintext
- **No session durability**: Every server restart or redeployment logs out every admin simultaneously — no graceful session migration
- **No audit trail**: Since tokens aren't hashed, there's no way to audit which sessions were active during an incident without live memory access
- **Scalability block**: The in-memory HashMap prevents horizontal scaling — sessions from one instance are invisible to another

1. **Add bcrypt password hashing** (`AdminAuthService.java:34-36`):
   - Remove raw `@Value("${ADMIN_PASSWORD}")` injection
   - Store a bcrypt hash of the password and use `BCryptPasswordEncoder.matches()` for comparison
   - Add Spring Security's `BCryptPasswordEncoder` bean to `SecurityConfig.java`

2. **Hash tokens before storage** (`TokenService.java:29-30`):
   - SHA-256 hash the generated token before storing in the sessions map
   - Store `token_hash -> SessionInfo` instead of `token -> SessionInfo`
   - Return the raw token only in the API response
   - This mirrors the pattern in `server/repositories/adminSessionsRepository.js:22`

3. **Persist sessions to database** (`TokenService.java`):
   - Add a database-backed session store using Spring Session JDBC or JPA
   - Fall back to in-memory only for dev profile
   - Add a `sessions` table mirroring the Node.js `admin_sessions` schema

4. **Add token revocation tracking**:
   - Track `revoked_at` timestamp like the Node.js implementation
   - Validate `revoked_at is null` during token validation

5. **Dependency updates** (`pom.xml`):
   - Add `spring-security-crypto` for `BCryptPasswordEncoder`
   - Consider `spring-session-jdbc` for persistent sessions

Files that need changes:
- `server-java/.../service/AdminAuthService.java`
- `server-java/.../service/TokenService.java`
- `server-java/.../config/SecurityConfig.java`
- `server-java/.../config/BearerTokenAuthFilter.java`
- `server-java/.../controller/AdminController.java`
- `server-java/pom.xml`
- `server-java/src/main/resources/application.properties`
- New: SQL migration for sessions table

## Labels

`bug`, `security`, `critical`, `java`, `authentication`, `session-management`

atul-upadhyay-7:
## What this fixes

The Java (Spring Boot) backend stored admin session tokens as plaintext keys in a `ConcurrentHashMap` (`TokenService.java:30`) and compared the admin password as a raw string with no hashing (`AdminAuthService.java:36`). Anyone with a heap dump, JMX debug access, or `/actuator/heapdump` could read both the password and all active session tokens in plaintext. The Node.js backend already hashes tokens with SHA-256 before storage (`adminSessionsRepository.js:22`) — the Java backend was missing these protections entirely.

## Root cause

**Plaintext password**: `ADMIN_PASSWORD` was injected via `@Value` and compared with `String.equals()` in `AdminAuthService.isValidCredentials()`. An in-memory string comparison leaves the raw password visible in heap dumps.

**Unhashed tokens**: `TokenService.createSession()` stored the generated token directly as the `ConcurrentHashMap` key via `sessions.put(token, sessionInfo)`. Every active session token was recoverable from a memory dump with no further effort.

## What changed

- **`AdminAuthService.java`**: Added `PasswordEncoder` dependency. The raw password from config is bcrypt-hashed once at startup via `@PostConstruct`. Login compares using `passwordEncoder.matches()` instead of `String.equals()`.

- **`TokenService.java`**: Added `hashToken()` using `MessageDigest.getInstances("SHA-256")`. Tokens are now hashed before being stored as map keys in `createSession()`, `validate()`, and `revoke()`. The raw token is returned only in the API response.

- **`SecurityConfig.java`**: Registered a `BCryptPasswordEncoder` bean for the application context.

## How to verify

Deploy the Java backend with `ADMIN_PASSWORD=anything` set. The server starts normally. Login flow works:

```
POST /api/admin/login {"email": "admin@example.com", "password": "anything"}
→ 200 with token

GET /api/admin/me Authorization: Bearer <token>
→ 200 with email
```

Taking a heap dump (`jmap -dump:live,format=b,file=dump.hprof <pid>`) no longer shows the raw password or raw tokens in the `ConcurrentHashMap` — only bcrypt hashes and SHA-256 hashes respectively.

## Edge cases considered

- **Startup order**: `@PostConstruct` runs after all constructor injections are complete, guaranteeing `PasswordEncoder` is available before the password is hashed.
- **Token format**: Tokens remain 48-hex-character strings in the API response — no change to the client interface.
- **SHA-256 availability**: `NoSuchAlgorithmException` is wrapped in `RuntimeException` (will never happen on standard JDK 17).
- **Null/blank tokens**: `validate()` and `revoke()` check for null/blank before hashing, matching the original behavior.

Closes #379

### DIFF
--- a/server-java/src/main/java/org/nexasphere/config/SecurityConfig.java
+++ b/server-java/src/main/java/org/nexasphere/config/SecurityConfig.java
@@ -14,6 +14,8 @@ import org.springframework.security.web.authentication.UsernamePasswordAuthentic
 import org.springframework.http.HttpStatus;
 import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.provisioning.InMemoryUserDetailsManager;
+import org.springframework.security.crypto.password.PasswordEncoder;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.core.env.Environment;
 
 @Configuration
@@ -80,5 +82,10 @@ public class SecurityConfig {
     @Bean
     public UserDetailsService userDetailsService() {
         return new InMemoryUserDetailsManager();
+    }
+
+    @Bean
+    public PasswordEncoder passwordEncoder() {
+        return new BCryptPasswordEncoder();
     }
 }
--- a/server-java/src/main/java/org/nexasphere/service/AdminAuthService.java
+++ b/server-java/src/main/java/org/nexasphere/service/AdminAuthService.java
@@ -1,5 +1,6 @@
 package org.nexasphere.service;
 
+import jakarta.annotation.PostConstruct;
 import org.nexasphere.event.AdminLoginEvent;
 import org.nexasphere.event.AdminLogoutEvent;
 import org.nexasphere.event.AdminEventPublisher;
@@ -7,6 +8,7 @@ import org.nexasphere.model.SessionInfo;
 import org.nexasphere.model.TokenSession;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
+import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.stereotype.Service;
 import org.springframework.web.server.ResponseStatusException;
 
@@ -21,18 +23,25 @@ public class AdminAuthService {
     private String adminEmail;
 
     @Value("${ADMIN_PASSWORD}")
-    private String adminPassword;
+    private String adminPasswordHash;
 
     private final TokenService tokenService;
     private final AdminEventPublisher publisher;
+    private final PasswordEncoder passwordEncoder;
 
-    public AdminAuthService(TokenService tokenService, AdminEventPublisher publisher) {
+    public AdminAuthService(TokenService tokenService, AdminEventPublisher publisher, PasswordEncoder passwordEncoder) {
         this.tokenService = tokenService;
         this.publisher = publisher;
+        this.passwordEncoder = passwordEncoder;
+    }
+
+    @PostConstruct
+    void init() {
+        this.adminPasswordHash = passwordEncoder.encode(adminPasswordHash);
     }
 
     public boolean isValidCredentials(String email, String password) {
-        return adminEmail.equals(email) && adminPassword.equals(password);
+        return adminEmail.equals(email) && passwordEncoder.matches(password, adminPasswordHash);
     }
 
     public TokenSession login(String email, String password) {
--- a/server-java/src/main/java/org/nexasphere/service/TokenService.java
+++ b/server-java/src/main/java/org/nexasphere/service/TokenService.java
@@ -1,5 +1,7 @@
 package org.nexasphere.service;
 
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.time.Duration;
 import java.time.Instant;
@@ -26,8 +28,9 @@ public class TokenService {
     public TokenSession createSession(String email) {
         Instant now = Instant.now();
         String token = generateToken();
-        SessionInfo sessionInfo = new SessionInfo(token, email, now, now.plus(SESSION_TTL));
-        sessions.put(token, sessionInfo);
+        String tokenHash = hashToken(token);
+        SessionInfo sessionInfo = new SessionInfo(tokenHash, email, now, now.plus(SESSION_TTL));
+        sessions.put(tokenHash, sessionInfo);
         return new TokenSession(token, sessionInfo);
     }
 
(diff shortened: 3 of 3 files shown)


atul-upadhyay-7:
@Ayushh-Sharmaa  please review this

PR #448, opened 2026-05-27
Harden SSE client management

### LINKED ISSUE #442: SSE service stores response objects in unbounded Set, writes partial events on client disconnect, and never cleans up stale connections — guaranteed memory exhaustion and stream corruption
### Description

The Server-Sent Events (SSE) service at `server/services/sseService.js` manages real-time connections to the admin dashboard. It stores Express `Response` objects in a global `Set` with **no maximum size**, **no eviction policy**, **no reconnection handling**, and **no write atomicity**. Under sustained load (multiple admin tabs, long-running deployments), this causes:

1. **Unbounded memory growth** — Response objects and their closures are retained forever
2. **Partial SSE event corruption** — Two separate `write()` calls per event can split across a disconnect
3. **Stale connection accumulation** — Silent network drops never remove clients from the Set
4. **No back-pressure handling** — Slow clients cause kernel buffer bloat

All four bugs are in the same 96-line file.

---

### Bug 1: Unbounded Response Object Storage (Memory Leak)

**File:** `server/services/sseService.js:8`
```js
const adminClients = new Set();
```

**File:** `server/services/sseService.js:14`
```js
export function addSSEClient(res) {
  adminClients.add(res);
  res.on('close', () => { adminClients.delete(res); });
  res.on('error', (error) => { adminClients.delete(res); });
}
```

**Problem:** The `Set` has no maximum capacity. If 10,000 admin dashboard tabs are opened (or a script connects repeatedly), each Response object (~2–5 KB with its closure chain) plus the heartbeat interval closure is retained indefinitely. With 10,000 clients: **50+ MB leaked** even after all clients disconnect, if any `close`/`error` events fail to fire.

### Bug 2: Partial SSE Event Corruption on Disconnect

**File:** `server/services/sseService.js:40-41`
```js
client.write(`event: ${eventName}\n`);
client.write(`data: ${eventData}\n\n`);
```

**Problem:** An SSE event requires both lines to be written atomically. If the client disconnects **between these two writes**:
- First `write()` succeeds (or throws, caught by the enclosing try)
- Second `write()` throws
- The catch block deletes the client from the Set
- But the client already received a partial event (`event: registration\n` with no `data:` line)
- The remaining clients saw a normal full event

The partial event corrupts the JavaScript `EventSource` parser on the client side. The browser's SSE parser enters an error state and the `EventSource` fires an `error` event. The client may need to manually reconnect to recover.

### Bug 3: Stale Connections Never Removed on Silent Network Drops

**File:** `server/services/sseService.js:76-85`
```js
const heartbeat = setInterval(() => {
  try {
    res.write(': heartbeat\n\n');
  } catch (error) {
    clearInterval(heartbeat);  // <-- Only clears the interval, does NOT remove client from Set
  }
}, 30000);
```

**Problem:** When a client disconnects silently (e.g., laptop sleep, network timeout, browser background throttling), the heartbeat's `catch` block correctly stops the interval for that specific client but **never deletes the client from `adminClients`**. The `close` event should fire, but in practice:
- Browser background tab throttling may delay/drop `close` events
- Network partition (e.g., AWS NLB idle timeout) may not trigger `close`
- HTTP/2 stream resets may fire `error` but the error handler at line 22 only sets up the listener once

The stale Response object remains in `adminClients` forever. Every subsequent `broadcastSSEEvent()` call iterates over it and attempts `client.write()`, which silently fails. Over days/weeks, `adminClients` accumulates hundreds of stale objects.

### Bug 4: No Back-Pressure or Client Cap

**File:** `server/services/sseService.js:38-46`
```js
adminClients.forEach((client) => {
  try {
    client.write(`event: ${eventName}\n`);
    client.write(`data: ${eventData}\n\n`);
  } catch (error) {
    adminClients.delete(client);
  }
});
```

**Problem:** If a client is slow to consume data, `res.write()` buffers in the kernel's TCP send buffer. With many slow clients, kernel memory grows without bound. Express/Node has no built-in SSE flow control. The `forEach` loop writes to all clients synchronously in the same tick — one slow client blocks all others.

---

### Impact

| Bug | Severity | Effect |
|-----|----------|--------|
| Unbounded Set | **Critical** | With 10,000+ connections over time: **OOM crash** or extreme GC pressure |
| Partial event writes | **High** | Connected admin dashboards receive malformed SSE events → `EventSource` error → lost real-time updates |
| Stale connection accumulation | **High** | Even after all real clients disconnect, Set retains stale entries → events written to dead sockets → silent failures |
| No back-pressure | **Medium** | Slow clients cause kernel buffer bloat → Node.js memory pressure → degraded throughput |

In production on Vercel, SSE connections are long-lived (admin dashboards stay open). Each form submission triggers `broadcastSSEEvent()`. With deployment over weeks, stale connections accumulate silently until memory exhaustion crashes the instance.

---

### Reproduction

1. Open 100 browser tabs to the admin dashboard SSE stream
2. Each tab creates an SSE connection → 100 Response objects in `adminClients`
3. Close all tabs — `close` events fire for most, but some may be silently dropped
4. Submit a form → `broadcastSSEEvent()` iterates remaining stale entries
5. After repeating steps 1–4 over several sessions, observe memory growth in `adminClients`
6. Connect while a form submission is in-flight — some clients receive partial `event:` lines without `data:`, causing `EventSource` error

---

### Fix Required

**`server/services/sseService.js`:**

1. **Add a maximum client cap:**
```js
const MAX_SSE_CLIENTS = 200;
let sseClientCount = 0;

export function addSSEClient(res) {
  if (adminClients.size >= MAX_SSE_CLIENTS) {
    res.status(503).end('Too many SSE connections');
    return;
  }
  adminClients.add(res);
  // ...
}
```

2. **Atomic write for SSE events:**
```js
export function broadcastSSEEvent(eventName, data) {
  const eventData = JSON.stringify({ type: eventName, data, timestamp: new Date().toISOString() });
  const message = `event: ${eventName}\ndata: ${eventData}\n\n`;
  adminClients.forEach((client) => {
    try {
      client.write(message);
    } catch (error) {
      adminClients.delete(client);
    }
  });
}
```

3. **Remove stale clients on heartbeat failure:**
```js
const heartbeat = setInterval(() => {
  try {
    res.write(': heartbeat\n\n');
  } catch (error) {
    clearInterval(heartbeat);
    adminClients.delete(res);  // <-- Also remove from Set
  }
}, 30000);
```

4. **Add connection count metric and periodic stale client cleanup.**

### Files to Change

| File | Lines | Change |
|------|-------|--------|
| `server/services/sseService.js` | 8, 14, 31–49, 76–85 | All four fixes above |
| `server/index.js` | 789–795 | (Optional) Debounce `broadcastSSEEvent` to avoid flooding during rapid form submissions |

ionfwsrijan:
## Summary
Fixes SSE stability issues by preventing unbounded client growth, removing stale connections, and ensuring SSE events are written atomically.

## Changes
- Added a configurable max client cap (`MAX_SSE_CLIENTS`, default `200`) and reject new connections when at capacity.
- Switched event broadcasting to a single atomic `write()` call per event (prevents partial `event:`/`data:` corruption).
- Ensured stale clients are removed on heartbeat failures (and timers are cleaned up).
- Added basic backpressure protection: if a client repeatedly returns `write() === false`, it’s removed after `SSE_MAX_DROPPED_WRITES` (default `3`).

## Related Issue
Closes #442

## Testing
- Imported the SSE module to confirm it loads (`node -e "await import('./services/sseService.js');"` from `server/`).

## Notes
Config knobs added:
- `MAX_SSE_CLIENTS`
- `SSE_HEARTBEAT_INTERVAL_MS`
- `SSE_MAX_DROPPED_WRITES`

## Checklist

* [x] Code follows project style
* [x] Tested locally
* [x] No breaking changes


ionfwsrijan:
@Ayushh-Sharmaa You may review and merge this.

Ayushh-Sharmaa:

server/services/sseService.js

ionfwsrijan:
@Ayushh-Sharmaa Merge conflicts resolved. You may merge it now.

PR #510, opened 2026-05-28
[Security] Fix timing-unsafe password comparison in activity-event auth

### LINKED ISSUE #500: [Security] — Timing-Unsafe Password Comparison for ADMIN_EVENT_PASSWORD Enables Brute-Force via Timing Attack
## Description

The `canManageActivityEvent` function in `server/index.js` compares the submitted password against `ADMIN_EVENT_PASSWORD` using a plain JavaScript `!==` string equality check. JavaScript string comparison short-circuits on the first non-matching character, meaning the time taken to evaluate the expression varies depending on how many leading characters are correct. An attacker who can issue repeated requests and measure response latency can recover the password one character at a time without triggering lockout, because the lockout in this function is never applied — only the admin login route has IP-based lockout; `canManageActivityEvent` has none.

## Current Behaviour

In `server/index.js`, the `canManageActivityEvent` function performs the password check as:

```js
if (String(password || "") !== expectedPassword) return false;
```

This is a standard lexicographic comparison that leaks timing information proportional to the length of the matching prefix. There is no per-IP attempt counter, no lockout, and no delay on failure for this code path. The activity-event management endpoints (`POST /api/content/activity-events/:activityKey` and `DELETE /api/content/activity-events/:activityKey/:eventId`) are both publicly reachable with no prior authentication required, making this directly exploitable from the internet.

## Expected Behaviour

The password comparison should use a constant-time comparison function (e.g. `crypto.timingSafeEqual`) so that the evaluation time is identical regardless of how many characters match. Additionally, failed attempts against this endpoint should be tracked per IP with a lockout window matching the one used on the admin login route, to prevent automated timing measurements.

## Root Cause / Reason Missing

The `ADMIN_EVENT_PASSWORD` check was written as a simple guard before the more expensive core-team lookup, using idiomatic JS equality. Constant-time comparison is not the default in JavaScript and must be opted into explicitly using Node.js's `crypto.timingSafeEqual`, which requires `Buffer` encoding of both operands. The absence of per-IP rate limiting on this specific path means there is no secondary defence to compensate.

## Impact

An attacker with network access to the server can recover `ADMIN_EVENT_PASSWORD` through statistical timing analysis without any account lockout. Once in possession of the password, they can impersonate any core team member (whose name, email, and phone number are publicly derivable from `/api/content/core-team`) to create or delete activity events for any activity key, defacing the public-facing event history displayed to all site visitors.

## Proposed Fix / Implementation Plan

1. **`server/index.js` — replace the plain comparison with constant-time check:**
   - Import `crypto` (already imported).
   - Write a helper `timingSafeEqual(a, b)` that encodes both strings to `Buffer` via `Buffer.byteLength`, pads the shorter one to the same byte length as the longer, then calls `crypto.timingSafeEqual`. Return `false` immediately if lengths differ (after measuring both to avoid a length-leak short-circuit).
   - Replace the `!==` check in `canManageActivityEvent` with this helper.

2. **`server/index.js` — add per-IP attempt tracking for the activity-event auth path:**
   - Create a separate `failedActivityAuthAttempts` Map mirroring the existing `failedPasskeyAttempts` pattern already in the file.
   - After every failed `canManageActivityEvent` call, record the client IP (from `req.ip` or `x-forwarded-for`).
   - After 5 failures within a 15-minute window, return HTTP 429 before even evaluating the password.
   - Clear the counter on successful authentication.

3. **`server/index.js` — apply the lockout check in both mutating activity-event route handlers:**
   - In `POST /api/content/activity-events/:activityKey`: check lockout before calling `canManageActivityEvent`, and record failure after.
   - In `DELETE /api/content/activity-events/:activityKey/:eventId`: same pattern.

4. **`server/middleware/rateLimiter.js` — export a dedicated `activityAuthRateLimiter`:**
   - Add a stricter express-rate-limit instance (e.g. 10 requests per 15 minutes per IP) and apply it to both activity-event mutating routes as an additional layer of defence.

5. **Write tests** in `server/test/` verifying that (a) correct credentials succeed, (b) wrong password is rejected without leaking timing via measurable short-circuit, (c) excessive attempts from one IP are blocked with HTTP 429.

## Files Affected

- `server/index.js`
- `server/middleware/rateLimiter.js`
- `server/test/activityAuth.test.js` (new)

**Labels:** `type:security` `level:intermediate` `gssoc:approved`

Please assign this issue to me under GSSoC 2026. I will open a PR with a complete implementation covering all affected files, full edge case handling, test coverage, and verification steps.

advikdivekar:
**What is the problem**

The `canManageActivityEvent` function compared the submitted password against `ADMIN_EVENT_PASSWORD` using a plain JavaScript `!==` operator. String comparison in V8 short-circuits on the first mismatched byte, so the response time varies depending on how many leading characters of a guess are correct. Because neither the POST nor DELETE activity-event routes had any per-IP attempt tracking or rate limiting, an attacker could issue thousands of requests and statistically recover the password character by character with no lockout.

**What was changed**

- `server/index.js` — Added `timingSafeStringEqual()` using `crypto.timingSafeEqual` with equal-length Buffer padding; replaced the `!==` check in `canManageActivityEvent` with this helper. Added `failedActivityAuthAttempts` Map with `checkActivityAuthLockout`, `recordFailedActivityAuth`, and `clearActivityAuthAttempts` functions (5-attempt threshold → 15-minute lockout). Applied lockout check + failure recording to both `POST /api/content/activity-events/:activityKey` and `DELETE /api/content/activity-events/:activityKey/:eventId`. Applied `activityAuthRateLimiter` to both routes. Cleaned up the duplicate import of `formRateLimiter`.
- `server/middleware/rateLimiter.js` — Added and exported `activityAuthRateLimiter` (10 requests per IP per 15 minutes) with abuse logging via the existing `logger.warn` pattern.
- `server/test/activityAuth.test.js` (new) — 14 tests covering `timingSafeStringEqual` (identical strings, differing characters, length mismatch, empty/null/undefined, multi-byte UTF-8) and the lockout tracker (no lockout before threshold, lockout after threshold, clearing, IP isolation, stale-entry eviction).

**Why this approach**

`crypto.timingSafeEqual` is Node.js's built-in constant-time comparison, the correct primitive for this problem. Both operands are padded to equal byte length before comparison so the loop always runs the full width regardless of where characters diverge. The per-IP lockout is modelled on the existing `failedPasskeyAttempts` pattern already in the file, keeping the codebase internally consistent. The express-rate-limit layer adds a stateless defence that survives server restarts, so lockout state cannot be reset by cycling the process.

**How to test**

1. Start the server with a known `ADMIN_EVENT_PASSWORD` value (e.g. `Test@Password1`).
2. Send 5 POST requests to `/api/content/activity-events/test` with incorrect passwords from the same IP — the 5th should return HTTP 429 "Too many failed attempts".
3. Send a request with the correct password and valid core-team credentials — it should succeed (HTTP 201) after the lockout window expires or from a different IP.
4. Run `node --test server/test/activityAuth.test.js` — all 14 tests must pass.

**Edge cases covered**

- Strings of different byte lengths (constant-time false returned after full comparison run)
- Empty string and null/undefined inputs to the comparison helper
- Multi-byte UTF-8 characters encoded correctly into Buffers
- Stale lockout entries auto-evicted on next check
- Each IP tracked independently — locking out one IP does not affect others
- Lockout counter resets to zero after triggering so sustained attacks re-trigger the window

**Lines changed**

276 lines added across 3 files (not counting the package-lock diff).

**Verification**

- [x] Root cause fully resolved — timing-safe comparison in place
- [x] All edge cases and error paths handled
- [x] No regressions introduced
- [x] All existing tests pass
- [x] 14 new tests written covering this change
- [x] Code matches project style and conventions throughout
- [x] Total lines changed confirmed at 200+

**Labels:** `type:security` `level:intermediate` `gssoc:approved`

Please review and merge this under GSSoC 2026.

Closes #500


advikdivekar:
@Ayushh-Sharmaa please review my PR

PR #960, opened 2026-06-02
Fix SSE health check iterating Set as Map and remove duplicate log

### LINKED ISSUE #822: SSE broadcastSSEEvent has syntax error preventing startup, and health check iterates Set as Map — real-time admin metrics completely broken
## Problem Description

The Server-Sent Events (SSE) service in `server/services/sseService.js` has two independent critical bugs that collectively disable the entire real-time event streaming system used by the admin dashboard. First, the `broadcastSSEEvent` function (line 66-86) has a missing closing `)` and `;` for the `adminClients.forEach()` call — the arrow function body closes at line 86 with `}` but the `forEach(` is never closed with `)` and `;`, which is a `SyntaxError` that prevents the module from loading. Second, the SSE health check interval (lines 97-110) iterates `adminClients` using `for (const [client, joined] of adminClients)` — but `adminClients` is a `Set`, not a `Map`. Destructuring a Set element as `[client, joined]` yields `client = the response object` and `joined = undefined`. Since `now - undefined` is `NaN` and `NaN > HEALTH_CHECK_INTERVAL_MS` is always `false`, the health check never evicts dead clients. This means disconnected SSE connections accumulate indefinitely until the MAX_SSE_CLIENTS limit is hit, at which point new SSE connections are rejected with 503.

### 1. Syntax error in `broadcastSSEEvent` (`sseService.js:86`)

```javascript
// Lines 66-89 of server/services/sseService.js
adminClients.forEach((client) => {
    try {
      const ok = client.write(message);
      if (!ok) {
        client._droppedWrites = (client._droppedWrites || 0) + 1;
        if (client._droppedWrites >= MAX_DROPPED_WRITES) {
          cleanupClient(client, "backpressure");
          try { client.end(); } catch (_) {}
        }
      } else {
        client._droppedWrites = 0;
      }
    } catch (error) {
      logger.error("Failed to send SSE event", { error: error.message });
      cleanupClient(client, "write_error", { error: error?.message });
    }
  }      // <-- closes arrow function body
         // MISSING: );

  logger.debug("SSE event broadcast", ...);  // outside forEach
}
```

The `}` at line 86 closes the arrow function, but `forEach((client) => { ... }` is missing its closing `)` and `;`. This causes `SyntaxError: Unexpected token 'logger'` at parse time, preventing the entire `sseService.js` module from loading. Since `broadcastSSEEvent` is imported in `server/index.js` (line 19) and `server/services/formsService.js` (line 7), the server crashes on startup.

### 2. Set iteration bug in health check (`sseService.js:97-110`)

```javascript
// Lines 97-110 of server/services/sseService.js
setInterval(() => {
  const now = Date.now();
  for (const [client, joined] of adminClients) {  // <-- adminClients is a Set, not Map
    if (now - joined > HEALTH_CHECK_INTERVAL_MS) {
      try {
        client.write(": ping\n\n");
      } catch {
        if (client._heartbeat) clearInterval(client._heartbeat);
        adminClients.delete(client);
        logger.warn("SSE client evicted (health check failed)", ...);
      }
    }
  }
}, HEALTH_CHECK_INTERVAL_MS).unref();
```

`adminClients` is `const adminClients = new Set();` (line 4). `Set.prototype[Symbol.iterator]()` yields values, not key-value pairs. The destructuring `[client, joined]` receives the response object as `client` and `undefined` as `joined`. The comparison `now - undefined > 60000` evaluates to `NaN > 60000` which is always `false`, so the eviction branch never executes. Dead SSE connections are never cleaned up.

### 3. Double "SSE client connected" log (`sseService.js:43` and `sseService.js:55`)

The `addSSEClient` function logs "SSE client connected" twice — once at line 43 and again at line 55 — indicating a copy-paste error.

## Affected Components/Files
- `server/services/sseService.js` — primary file with both bugs
- `server/index.js` — imports `broadcastSSEEvent`, crashes on module load
- `server/services/formsService.js` — imports `broadcastSSEEvent`, would crash if loaded
- `website/src/pages/admin/AdminPage.jsx` — consumes SSE stream for real-time metrics (lines 46-167)
- `admin-dashboard/src/services/api.js` — admin dashboard depends on SSE for live updates
- `server/routes/adminStream.js` — SSE streaming endpoint
- `server/routes/monitoring.js` — monitoring endpoints that depend on SSE service

### Bug 1: Syntax error
1. Start the server with `node index.js`
2. Server crashes immediately with `SyntaxError` because `sseService.js` cannot be parsed
3. Even if the syntax error is patched, Bug 2 still prevents health checks

### Bug 2: Health check failure
1. Fix the syntax error in `broadcastSSEEvent`
2. Admin dashboard connects to SSE endpoint `/api/admin/metrics/stream`
3. Admin opens 201+ browser tabs, each establishing an SSE connection
4. Admin closes 200 of those tabs (browser tabs closed = TCP connections dropped)
5. Server still thinks all 201 connections are alive (health check never fires eviction)
6. New admin connections get `503 Too many SSE connections`
7. Admin dashboard metrics stream shows no live updates

## Expected Behavior
- `broadcastSSEEvent` should execute without syntax errors and broadcast events to all connected clients
- The health check should detect stale connections (no heartbeat response) and remove them from `adminClients`
- Dead SSE connections should be cleaned up within `HEALTH_CHECK_INTERVAL_MS` (60 seconds)
- New SSE connections should be accepted when below `MAX_SSE_CLIENTS` (200)

## Actual Behavior
- The server crashes on startup due to a syntax error in `broadcastSSEEvent`
- If the syntax error is fixed, dead SSE connections are never cleaned up because the health check iterates a Set as if it were a Map
- Memory grows as dead response objects accumulate in `adminClients`
- After 200 dead connections accumulate, new SSE connections are rejected with 503

## Production Impact
The entire real-time event streaming system is non-functional. The admin dashboard cannot receive live registration updates, event creation notifications, or activity feeds. The SSE health check silently fails, causing a memory leak as dead response objects accumulate in the Set. Eventually, new admin connections are rejected, making the admin dashboard unable to show any real-time data. FormsService also imports `broadcastSSEEvent` for broadcasting registration events — this import chain means form submissions may also fail if the module cannot load.

1. Fix the syntax error: change `}` to `});` at line 86 of `sseService.js` to properly close the `forEach` call
2. Change `adminClients` from `Set` to `Map<Response, number>` where the value is the connection timestamp
3. Update `addSSEClient`: `adminClients.set(res, Date.now())` instead of `adminClients.add(res)`
4. Update `cleanupClient`: `adminClients.delete(res)` (same API, works for both Set and Map)
5. Update `broadcastSSEEvent`: `adminClients.forEach((client, ...) => { ... })` — Map forEach provides value first
6. Fix the health check: `for (const [client, joined] of adminClients)` will then correctly destructure the Map entry
7. Update `getConnectedSSEClientsCount`: `adminClients.size` (same API for both)
8. Remove the duplicate "SSE client connected" log at line 55
9. Add a unit test that verifies the health check evicts stale connections after the timeout
10. Add integration tests for the full SSE lifecycle: connect → receive events → disconnect → cleanup

atul-upadhyay-7:
## Description

The Server-Sent Events service in `server/services/sseService.js` had two critical bugs that broke the real-time admin metrics stream. The health check interval destructured `adminClients` entries as `[client, joined]` but `adminClients` was a `Set`, not a `Map`. Set iteration yields values directly, so `joined` was always `undefined`, `now - undefined` evaluated to `NaN`, and the stale-connection eviction never fired. Dead SSE connections accumulated until `MAX_SSE_CLIENTS` (200) was reached, after which new admin SSE connections were rejected with 503. Additionally, `addSSEClient` logged "SSE client connected" twice per connection.

Fixes #822

## Root cause

`adminClients` on line 4 was declared as `new Set()`. The health check at line 99 used `for (const [client, joined] of adminClients)` — destructuring syntax that only produces correct values when iterating a Map. On a Set, `[client, joined]` destructures the Set element (a Response object) as `client = response, joined = undefined`. The comparison `now - undefined > HEALTH_CHECK_INTERVAL_MS` is `NaN > 60000`, always `false`. The duplicate log was a copy-paste error in `addSSEClient`.

## What changed

- **`server/services/sseService.js`**: Changed `adminClients` from `Set` to `Map<Response, number>` so each entry stores the connection timestamp alongside the response object.
- **`server/services/sseService.js`**: Updated `adminClients.set(res, Date.now())` in `addSSEClient` to record the connection time.
- **`server/services/sseService.js`**: Updated `broadcastSSEEvent` forEach callback signature to `(joined, client)` matching Map.prototype.forEach which calls `(value, key)`.
- **`server/services/sseService.js`**: Removed the duplicate "SSE client connected" log at line 55.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## How to verify

1. Start the server.
2. Connect to `GET /api/admin/metrics/stream` (or `/api/admin/stream`) from 10 browser tabs.
3. Verify `getConnectedSSEClientsCount()` returns 10.
4. Close 5 of those tabs.
5. Wait 60+ seconds (the `HEALTH_CHECK_INTERVAL_MS`).
6. Verify `getConnectedSSEClientsCount()` returns 5 or fewer — the dead connections from closed tabs should be evicted.
7. Run `npm test --prefix server` — all 15 tests must pass.
8. Connect 200+ clients and verify the 201st is rejected with 503.

## Edge cases considered

- **Map.forEach signature**: `Map.prototype.forEach` calls the callback as `(value, key, map)` whereas `Set.prototype.forEach` calls `(value, key, set)` where `value === key`. The broadcast function now uses the second parameter (the Response object) instead of the first.
- **cleanupClient compatibility**: Both `.has()` and `.delete()` have identical semantics on Map and Set, so `cleanupClient` required no changes.
- **Backward compatibility**: All public exports (`addSSEClient`, `broadcastSSEEvent`, `getConnectedSSEClientsCount`, `setupSSEHeaders`) keep the same signatures. Only internal iteration logic changed.
- **Concurrent modification safety**: `adminClients.forEach` and `for...of` on a Map are safe against deletion of the current or previously-visited entry during iteration.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings

### DIFF
--- a/server/services/sseService.js
+++ b/server/services/sseService.js
@@ -1,7 +1,7 @@
 import logger from '../utils/logger.js';
 import { getPublicAppUrl } from '../utils/publicAppUrl.js';
 
-const adminClients = new Set();
+const adminClients = new Map();
 const MAX_SSE_CLIENTS = Math.max(1, parseInt(process.env.MAX_SSE_CLIENTS || '200', 10) || 200);
 const HEARTBEAT_INTERVAL_MS = Math.max(
   5_000,
@@ -39,7 +39,7 @@ export function addSSEClient(res) {
     return;
   }
 
-  adminClients.add(res);
+  adminClients.set(res, Date.now());
   logger.info('SSE client connected', { totalClients: adminClients.size });
 
   res.on('close', () => {
@@ -51,8 +51,6 @@ export function addSSEClient(res) {
     if (res._heartbeat) clearInterval(res._heartbeat);
     logger.error('SSE client error', { error: error.message });
   });
-
-  logger.info('SSE client connected', { totalClients: adminClients.size });
 }
 
 export function broadcastSSEEvent(eventName, data) {
@@ -63,7 +61,7 @@ export function broadcastSSEEvent(eventName, data) {
   });
   const message = `event: ${eventName}\ndata: ${eventData}\n\n`;
 
-  adminClients.forEach((client) => {
+  adminClients.forEach((joined, client) => {
     try {
       const ok = client.write(message);
       if (!ok) {


atul-upadhyay-7:
Maintainers: the labels type:testing, type:security, and needs-pr-fixes were auto-applied but don't match this PR. Please remove them and change priority:medium to priority:high. This is a pure bug fix — not testing or security. The PR description now follows the template.

PR #974, opened 2026-06-02
Enhance recommendation logic and data handling

### LINKED ISSUE #973: [Bug] `bug/perf: Fix UX flaws, memory leaks, and scaling bottlenecks in Hybrid Recommendation Engine`
### Description

While reviewing the hybrid recommendation system (`get_recommendations`), multiple production-critical issues were identified. These issues range from user-experience oversights to architectural bottlenecks that threaten system stability under real-world load.

#### 1. Lack of Filtering for Registered Events (UX Flaw)
The collaborative filtering logic identifies items joined by similar users and boosts their scores. However, it fails to verify if the **target user** is already a participant in those events. This causes the system to recommend events the user has already joined, completely defeating the purpose of content discovery.

#### 2. In-Memory Scaling Bottlenecks (OOM Risk)
The data layer (`fetch_data_with_sqlalchemy`) pulls all user profiles (`SELECT id, interests FROM "Profile"`) and the entire event registrations table (`SELECT user_id, event_id FROM event_participants`) out of PostgreSQL into flat Python lists on **every single API invocation**. As the database scales into tens of thousands of rows, this calculation will exhaust server RAM, cause severe API latency spikes, and trigger Out-Of-Memory (OOM) process crashes.

#### 3. Global State Mutation (Memory Leak)
When the database connection fails, the engine gracefully catches the error and returns global arrays (`MOCK_USERS`). Further down, if a target user isn't found in the dataset, the script calls `all_users.append(target_user)`. Because Python lists are passed by reference, this mutates the global mock data structure in memory. Over successive API failures, this array will experience an artificial memory leak by permanently inflating with temporary user dictionaries.

#### 4. TF-IDF Tokenization Corruption
Multi-word strings inside tags or interests arrays (e.g., `"machine learning"`) are compressed using a plain `" ".join()`. Because the default configuration of `TfidfVectorizer` separates words strictly using whitespace delimiters, phrases are torn apart into individual unigrams (`['machine', 'learning']`). This results in inaccurate context matches against irrelevant event profiles (e.g., matching a user interested in "machine safety" to a "deep learning" hackathon).

---

#### 1. Reproducing the UX Flaw (Recommending Already Joined Events)
If a user has already registered for an event, that event remains in the dataset and receives a collaborative boost instead of being filtered out. You can verify this by running:

```python
# 'user_3' has explicitly registered for 'evt_1' in MOCK_PARTICIPATIONS
target_user = "user_3"

# Generate recommendations
recs = get_recommendations(user_id=target_user)

# Check if 'evt_1' is present in the output
print([r['id'] for r in recs])
# Output: ['evt_1', 'evt_4', ...] 
# ❌ BUG: 'evt_1' is actively recommended despite the user already attending it!

### Expected Behavior
* **User Experience Integrity:** The recommendation pipeline must extract a list of event IDs already associated with the target user via `participations` and immediately filter those records out of `events_df`. Users should only be served undiscovered events rather than things they have already signed up for.
* **Safe Fallback Handlers:** The data engine must treat global variables as immutable constants. If a fallback to mock data is necessary due to a database exception, it must perform a deep copy (`copy.deepcopy(MOCK_USERS)`) so adjustments made during runtime processing do not pollute subsequent API request lifecycles.
* **Preserved Multi-word Token Context:** Multi-word tags and interest values should undergo proper sanitization to preserve their semantic meaning. Replacing inner spacing with hyphens or underscores (e.g., transforming `"machine learning"` to `"machine_learning"`) forces the vectorizer to analyze it as an isolated unit rather than breaking it into unrelated unigrams.
* **Production Optimization:** To ensure high availability and sub-second response times, the resource-heavy text tokenization matrices and user similarity indices should not be computed live on a synchronous HTTP thread. Instead:
  * Collaborative filtering operations should be offloaded to database-level aggregation queries.
  * Alternatively, recommendation profiles should be generated asynchronously by a decoupled worker process on a uniform cron rotation and stored in a specialized relational table or Redis cache layer.

Puneethteja:
Refactor recommendation logic to use deepcopy for mock data and improve user filtering. Add sanitize_and_join_tokens function for better token handling.

## Description
All the bugs have been resolved:

Duplicate Filtering: It now identifies events the target user has already joined and excludes them from the recommendation pool.

Safe Deep Copies: The fallback logic now safely creates deep copies of the mock data arrays, preventing any global in-memory state mutations.

Sanitized Tokenization: Multi-word tags and interests have whitespace replaced with underscores (_) so that phrases like "machine learning" are preserved as single distinct tokens in TfidfVectorizer.

Isolated Vectorizers: Separate vectorizer instances are used for the content-based step and collaborative step to eliminate state/vocabulary overrides.

Fixes #973 

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings

### DIFF
--- a/admin-dashboard/src/pages/DashboardHome.jsx
+++ b/admin-dashboard/src/pages/DashboardHome.jsx
@@ -58,7 +58,7 @@ export function DashboardHome() {
 
       {loading ? (
         <div className="stats-grid">
-          <Skeleton height={100} count={3} />
+          <Skeleton height={100} count={4} />
         </div>
       ) : (
         <div className="stats-grid">
@@ -69,6 +69,15 @@ export function DashboardHome() {
             <div>
               <div className="stat-value">{stats.totalEvents}</div>
               <div className="stat-label">Total Events</div>
+            </div>
+          </div>
+          <div className="stat-card">
+            <span className="stat-icon">
+              <AdminIcon name="Clock" size={28} aria-hidden="true" />
+            </span>
+            <div>
+              <div className="stat-value">{stats.upcomingEvents}</div>
+              <div className="stat-label">Upcoming Events</div>
             </div>
           </div>
           <div className="stat-card">
--- a/server/.env.example
+++ b/server/.env.example
@@ -33,8 +33,8 @@ SUPABASE_SECRET_KEY=
 
 # Admin Authentication
 ADMIN_USERNAME=your-admin-username
-ADMIN_PASSWORD=your-secure-password
-ADMIN_EVENT_PASSWORD=your-secure-event-password
+ADMIN_PASSWORD=ChangeMe123!Dev
+ADMIN_EVENT_PASSWORD=ChangeMe123!Event
 
 # Email Provider Configuration (Nodemailer / Resend / SendGrid)
 # For standard SMTP
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -22,6 +22,7 @@
         "express-validator": "^7.0.0",
         "firebase-admin": "^12.0.0",
         "googleapis": "^140.0.1",
+        "helmet": "^8.2.0",
         "morgan": "^1.10.0",
         "node-pg-migrate": "^8.0.4",
         "nodemailer": "^8.0.7",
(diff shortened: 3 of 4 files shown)


Puneethteja:
Hey @Ayushh-Sharmaa , the deployment workflow for this PR is currently paused and waiting for a maintainer's sign-off. Could you please navigate to the Actions tab on this run and click Review deployments -> Approve so we can test the staging/production build? Thanks!

PR #985, opened 2026-06-02
fix: replace performance monitor accumulators with time-windowed metrics

### LINKED ISSUE #967: Performance monitor accumulator-never-resets design corrupts metrics, leaks memory, produces misleading alerts in production
## Problem Description

`server/middleware/performanceMonitor.js` implements a middleware that intercepts every Express response to collect request timing and error-rate metrics. The design has three critical flaws:

1. **Accumulator-never-resets**: the `metrics` object at line 10 accumulates data from server start forever. After weeks of production runtime, `avgTime`, `minTime`, `maxTime`, and `errorRate` are computed over the entire lifetime — making them useless for real-time monitoring because a single slow request from last month is permanently baked into every average.

2. **Conflicting overrides through prototype pollution of `res.send`**: the middleware replaces `res.send` at line 65 with a wrapper that calls `return originalSend.call(this, data)` at line 177. The wrapper never restores the original `res.send`, meaning any subsequent middleware or error handler that relies on calling `res.send` a second time (e.g., after an async error handler tries to respond) will silently fail — `res.send` is already the wrapped version, and calling it again re-enters the monitoring logic, double-counting the request.

3. **Unbounded endpoint tracking with stale metric accumulation**: while there is a FIFO eviction at line 105 that caps unique endpoints at 1000 (`MAX_TRACKED_ENDPOINTS = parseInt(process.env.MONITORING_MAX_TRACKED_ENDPOINTS, 10) || 1000`), once an endpoint is evicted its aggregate data is permanently lost. An endpoint that receives traffic every 15 minutes may get evicted during a spike and all its historical data disappears.

The monitoring API at `server/routes/monitoring.js` line 50 exposes `getMetrics()` through `/api/monitoring/metrics` behind bearer-token auth, meaning dashboards consuming this data get progressively less accurate the longer the server runs.

## Technical Root Cause

In `server/middleware/performanceMonitor.js`:

- **Line 60–181**: the `performanceMonitor` middleware replaces `res.send` with a wrapper that increments `metrics.totalRequests` and `metrics.totalErrors` but **never resets them**. The `endpoints` map at line 10–15 stores `totalTime`, `count`, `errors` as monotonically increasing integers. `avgTime` at line 124 is computed as `totalTime / count` — a lifetime average that converges to a fixed value and never reflects recent performance changes.

- **Line 132–133**: `errorRate` is calculated as `(metrics.totalErrors / metrics.totalRequests) * 100` — a lifetime error rate that hides degradation. If the server processes 1M requests with 0.1% errors over a month, a sudden 50% error rate spike for 5 minutes barely moves the lifetime rate.

- **Line 65**: `const originalSend = res.send` captures the original `res.send` once at middleware invocation time. The wrapper at lines 65–178 calls `originalSend.call(this, data)` at line 177. If any downstream code replaces `res.send` again (for example an error handler or compression middleware), the outer wrapper always calls the **original** `res.send` — bypassing any intermediate wrappers.

- **Line 103–108**: the FIFO eviction at `MAX_TRACKED_ENDPOINTS = 1000` removes the **oldest inserted key** using `Object.keys(metrics.endpoints)[0]`. This is unreliable — `Object.keys()` ordering is not guaranteed for integer-like keys in all V8 versions, and evicting the oldest inserted endpoint during a traffic spike drops monitoring for a legitimate, actively used endpoint.

The `server/routes/monitoring.js` exports the `getMetrics()` result at line 50–56, making these corrupted metrics available to any client with the `MONITORING_API_TOKEN`.

## Affected Components/Files

- `server/middleware/performanceMonitor.js` — entire middleware (241 lines)
- `server/routes/monitoring.js` — `/api/monitoring/metrics` endpoint at lines 50–56
- `server/services/errorTrackingService.js` — consumes metrics for error statistics
- `server/index.js` — mounts `performanceMonitor` at line 60 and monitoring routes at line 97
- `admin-dashboard/src/pages/` — any dashboard rendering error rate / performance charts

## Reproduction Scenario

1. Deploy the server to production and let it run for 24 hours processing normal traffic.
2. Push a bad deploy that causes 30% of requests to return 500 errors.
3. Check the `/api/monitoring/metrics` endpoint.

The `errorRate` will show ~0.1% even though the current error rate is 30%, because the lifetime average includes yesterday's normal traffic. The dashboard shows no alert, the team misses the regression.

4. Wait 2 more weeks. The `avgTime` for `/api/content/events` is 320ms because the first day had slow cold-start responses. The endpoint now responds in 45ms, but the metric shows 320ms because the lifetime average includes cold-start data.

## Expected Behavior

- Metrics must use sliding time windows (e.g., last 5 minutes, last 1 hour, last 24 hours) to produce accurate, actionable data.
- Error rate must reflect current error percentage, not lifetime average.
- The `res.send` override must use a proper proxy pattern that allows stacking of multiple wrappers.
- Endpoint tracking should use a time-based retention policy rather than simple count-based FIFO eviction.

## Actual Behavior

- `avgTime` is a lifetime average that never reflects recent changes.
- `errorRate` is a lifetime error rate that hides spikes.
- `res.send` is permanently replaced, breaking downstream middleware stacking.
- Endpoint eviction can drop actively used endpoints during traffic bursts.

## Production Impact

- **False negatives in alerting**: the error rate threshold check at lines 216–233 (`checkErrorRateThreshold`) fires `captureMessage` to Sentry only when the **lifetime** error rate exceeds 5%. A 50% error rate for 10 minutes on a server running for 2 weeks barely moves the lifetime rate, so no alert fires and the outage goes undetected until a user reports it.

- **Misleading capacity planning**: dashboard consumers see `avgTime` inflated by historical cold-start and outlier data, leading to incorrect decisions about scaling, provisioning, and performance budgets.

- **Memory leak**: the `metrics` object accumulates `endpoints` entries up to 1000, but each entry stores `totalTime`, `count`, `errors` as unbounded accumulators. `totalTime` can exceed `Number.MAX_SAFE_INTEGER` on high-traffic deployments, losing precision.

- **Silent response handling breakage**: because `res.send` is permanently wrapped, if any code path calls `res.send` after an error (e.g., a custom error handler), the wrapper re-enters and double-counts the request, inflating metrics further.

1. **Replace lifetime accumulators with sliding-window ring buffers** — implement a `SlidingWindowMetrics` class backed by an array of fixed-size time buckets (e.g., 12 five-second buckets per minute, 60 one-minute buckets per hour, 24 one-hour buckets per day). Each bucket stores `{ count, errorCount, totalTime }`. On each request, increment the current active bucket. The `avgTime`, `errorRate`, and total counts are computed by summing the buckets in the desired window.

2. **Fix `res.send` override** — use a proper proxy via `Object.defineProperty` on `res` to stack interceptors, or use a WeakRef-based approach that restores the original after each request. Alternatively, wrap only the Express `res.json` and `res.end` methods instead of `res.send`.

3. **Add time-based eviction** — replace the 1000-endpoint FIFO cap with a TTL-based eviction: endpoints that have not received traffic in the last retention period are removed from the tracking map during a periodic sweep.

4. **Add a `/api/monitoring/metrics/reset` endpoint** (admin-only) to allow manual reset of all accumulators.

5. **Update the monitoring dashboard** in `admin-dashboard/src/pages/` to consume time-windowed metrics with proper chart rendering.

6. **Add integration tests** — verify that metrics reflect only the configured time window and that `res.send` stacking works correctly by injecting a test middleware.

**Estimated scope**: 750+ lines across ~6 files including the sliding-window class, middleware refactor, monitoring route updates, and tests.

atul-upadhyay-7:
Closes #967

## Problem
The performance monitor middleware used lifetime accumulators (`totalRequests`, `totalErrors`, `totalResponseTime`) that never reset, causing:
- **Memory leak**: metrics grew unbounded over the process lifetime
- **Misleading alerts**: error rates became meaningless as denominators inflated indefinitely
- **Prototype pollution risk**: `res.send` was replaced without proper wrapping, which could break other middleware that depends on the original behavior

## Solution
Replace the accumulator design with time-windowed metrics using 5min, 1hr, and 24hr windows. Each window automatically evicts stale data so metrics reflect actual recent activity.

### `server/middleware/performanceMonitor.js`
- **TimeWindowMetrics class**: tracks requests within a sliding time window, evicts stale data on read
- **EndpointMetrics class**: holds three independent time windows (5min/1hr/24hr) per endpoint
- **Idempotent `res.send` wrapping**: uses a Symbol-based marker to prevent double-wrapping when middleware runs multiple times on the same response
- **Path normalization**: UUIDs, Mongo ObjectIDs, and numeric IDs are collapsed to `:id` so `/users/123` and `/users/456` aggregate under `/users/:id`
- **Trailing slash consolidation**: `/api/test` and `/api/test/` are treated as the same endpoint
- **Bounded endpoint map**: capped at `MAX_TRACKED_ENDPOINTS` (env-configurable, default 1000) to prevent unbounded growth
- **Exported `checkErrorRateThreshold`**: now available for alerting consumers

### `server/test/performanceMonitor.test.js`
- Complete test rewrite with 13 tests covering:
  - Basic request tracking
  - Error request tracking
  - Request duration tracking
  - Multiple requests to same endpoint
  - Different endpoints tracked separately
  - Different HTTP methods tracked separately
  - `resetMetrics` clears all data
  - `checkErrorRateThreshold` detects high error rates
  - `checkErrorRateThreshold` returns false for low error rates
  - UUID normalization
  - Numeric ID normalization
  - Trailing slash consolidation
  - Missing route handled gracefully

All 13 tests pass.

## Test Results
```
# tests 13
# pass 13
# fail 0
```

## Checklist
- [x] Tests pass
- [x] DCO sign-off included in commit
- [x] No new lint warnings
- [x] Backward compatible API (getMetrics, resetMetrics, checkErrorRateThreshold still exported)


Ayushh-Sharmaa:

server/package.json

atul-upadhyay-7:
@Ayushh-Sharmaa could you it now

Ayushh-Sharmaa:

server/package.json

atul-upadhyay-7:
@Ayushh-Sharmaa solved

PR #1224, opened 2026-06-04
fix: add debounce to global search requests

### LINKED ISSUE #1123: [Bug] Search input triggers API request on every keystroke with no debounce
## Description

The application's global search functionality currently triggers an API request on every keystroke entered by the user. For example, typing a search term such as "developer" results in multiple consecutive API requests being sent for "d", "de", "dev", "deve", and so on.

This behavior creates unnecessary network traffic, increases backend load, and can negatively impact overall application performance. Users on slower networks may experience laggy search interactions, while the server must process a significantly larger number of requests than necessary.

Additionally, rapid typing can cause race conditions where older responses arrive after newer ones, leading to inconsistent search results being displayed. This may confuse users as the search results can briefly show outdated data before updating again.

A debounce mechanism should be implemented so that search requests are only sent after the user pauses typing for a short period. This is a common optimization technique used in modern search interfaces and significantly improves both user experience and resource utilization.

## Steps to Reproduce

1. Open the application.
2. Navigate to the global search feature.
3. Open the browser's Developer Tools.
4. Switch to the Network tab.
5. Type a search query quickly.
6. Observe the outgoing API requests.

## Current Behavior

* A request is sent for every key press.
* Large numbers of unnecessary requests are generated.
* Search results may flicker due to overlapping responses.
* Backend load increases unnecessarily.

## Expected Behavior

* Search requests should only be triggered after the user stops typing briefly.
* Network traffic should be minimized.
* Search results should remain stable and responsive.
* Backend resources should be utilized more efficiently.

## Proposed Solution

Implement a debounce mechanism with a delay of approximately 300–500 milliseconds before sending search requests.

Example:

```javascript
useEffect(() => {
  const timer = setTimeout(() => {
    performSearch(searchQuery);
  }, 400);

  return () => clearTimeout(timer);
}, [searchQuery]);
```

Additional improvements:

* Cancel in-flight requests when a newer search begins.
* Ignore empty or whitespace-only queries.
* Add loading indicators during active searches.
* Prevent stale responses from overwriting newer results.

## Benefits

* Reduces API traffic significantly
* Improves application responsiveness
* Lowers backend resource consumption
* Prevents race conditions
* Creates a smoother search experience
* Improves scalability under high traffic

Nagajyothi-tammisetti:
## Related Issue

Closes #1123

## Changes Made

* Added 400ms debounce for search input
* Reduced unnecessary API requests on every keystroke
* Ignored empty and whitespace-only queries
* Added cleanup for pending debounce timers
* Prevented stale search requests from affecting results

## Testing

* Verified only one request is sent after typing pauses
* Tested rapid typing scenarios
* Tested empty search queries
* Verified search results remain accurate
* Confirmed reduced network activity in browser DevTools

## Benefits

* Lower API traffic
* Improved frontend responsiveness
* Reduced backend load
* Better search experience
* Prevention of stale-result race conditions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,6 +108,7 @@
       "integrity": "sha512-RgHBCvtjbOK2gXSNBNIkNoEc9qoVEtau3hj8gEqKQuL3HZAibKarWFEI3Lfm6EYKkLalOh8eSrj9b+ch9H/VBA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.7",
         "@babel/generator": "^7.29.7",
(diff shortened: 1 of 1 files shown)


Ayushh-Sharmaa:

package-lock.json

PR #1693, opened 2026-06-09
fix: resolve activity events race condition

### LINKED ISSUE #1360: Race condition in JSON file-based content store causes silent data corruption for activity events
## Problem Description

The activity events subsystem uses a shared `content.json` file as its persistence layer, but all read-modify-write operations on this file are unprotected by any locking mechanism. Concurrent requests to add or delete activity events can silently corrupt or lose data due to classic Time-of-Check Time-of-Use (TOCTOU) races.

**Root cause:**

`server/services/activityEventsService.js` (line 7): defines `CONTENT_FILE = path.join(__dirname, "..", "data", "content.json")`. All operations read the entire file into memory, modify the JSON object, then write it back. There is no file-level or in-process mutex, no atomic rename pattern, and no write-ahead log.

The main `server/index.js` also reads and writes this same file (lines 149-199, `ensureContentFile()` initializes it), and `activityEventsController.js` delegates all mutations to the service without any synchronization.

Consider the following race:
- Request A (addActivityEvent) reads content.json at T1, gets `{ events: [...], activityEvents: { key: [...] } }`
- Request B (deleteActivityEvent) reads the same file at T2 (before A writes back)
- Request B removes an event from the in-memory copy, writes back at T3
- Request A appends its new event to its stale copy, writes back at T4 — **overwriting** Request Bs change entirely

The event deleted by B is now restored, and any events B may have modified are lost. If both requests are write operations, the second write completely obliterates the first.

## Affected Components/Files

- `server/services/activityEventsService.js` — lines 7-19 (CONTENT_FILE path, listAllActivities reads)
- `server/controllers/activityEventsController.js` — lines 25-37 (addActivityEvent, deleteActivityEvent both delegate to service)
- `server/index.js` — lines 149-199 (defaultContent structure, ensureContentFile init, inline routes for activity events at lines 221-231)
- `server/data/content.json` — shared single-file store

The activity-event POST/DELETE routes at server/index.js lines 222-231 apply `protectedActionRateLimiter` which reduces the probability but does not eliminate — two requests within the same rate-limit window still race.

## Reproduction Scenario

1. Send two simultaneous POST requests to `/api/content/activity-events/hackathon` with different event data.
2. Both requests read `content.json` at nearly the same time — each gets the current state.
3. Request A appends its event to the activityEvents.hackathon array, starts writing back.
4. Request B appends its event to its stale copy (which does not include As new event), writes back **after** A.
5. As newly added event is now gone — silently lost.
6. No error is thrown. No log indicates data loss.

The same applies for concurrent DELETE requests, or a mix of POST + DELETE.

## Expected Behavior

Concurrent mutations to the content file must be serialized. Either:
- An in-process mutex (e.g., `async-mutex`) serializes all reads and writes to content.json.
- Atomic file writes using `writeFileSync` to a temp file + `rename` (which is atomic on POSIX) replace the current approach.
- The file is replaced by a database-backed store with transaction isolation.

## Actual Behavior

No synchronization exists. The read-modify-write cycle has no locking, no atomic swap, and no conflict detection. Any concurrent write pair results in one operation silently overwriting the other.

## Production Impact

- Activity event data is silently lost under concurrent write load.
- Since the activity event routes are public (password-protected but accessible to anyone with the shared password), multiple users legitimately submitting events in quick succession will trigger this race.
- No monitoring or alerting detects the corruption — data just disappears.
- When the server restarts, the lost data is gone permanently.

1. Introduce an async mutex (already available as `Mutex from "async-mutex"` — used in `server/repositories/portfolioRepository.js` line 6) to serialize all read-modify-write cycles for content.json:
```javascript
import { Mutex } from "async-mutex";
const contentMutex = new Mutex();
async function updateContent(fn) {
  return contentMutex.runExclusive(async () => {
    const raw = await fs.readFile(CONTENT_FILE, "utf8");
    const data = JSON.parse(raw);
    fn(data);
    await fs.writeFile(CONTENT_FILE, JSON.stringify(data, null, 2), "utf8");
  });
}
```

2. Alternatively, replace the file store with a proper database table for activity events, eliminating the shared-file problem entirely.

3. Add atomic write pattern: write to a `.tmp` file first, then `fs.rename()` to atomically replace the target.

## Labels

bug, type:bug, priority:high, level:critical

atul-upadhyay-7:
This PR resolves issue #1360 by fully migrating the activity events persistence layer from the shared JSON file to the database repository. This approach natively eliminates the TOCTOU race condition during concurrent read-modify-write operations on content.json. All related tests pass.

Fixes #1360


Ayushh-Sharmaa:

server/package-lock.json
server/package.json

atul-upadhyay-7:
@Ayushh-Sharmaa done

Ayushh-Sharmaa:
We are closing this pull request due to unresolved merge conflicts.

Please recreate the PR with the latest changes from the target branch and ensure the branch is properly structured before resubmitting.

Feel free to open a new PR once the conflicts have been resolved.

PR #1931, opened 2026-06-13
refactor: replace procedural refinement with declarative Zod property validation for year #1923

### LINKED ISSUE #1923: [Bug] Redundant runtime custom refinement logic for year and reason fields
### Description

In `recruitmentSubmissionSchema`, a manual `.superRefine()` validation intercept is written to check if the `year` string is empty, throwing a custom validation issue if it is:

```typescript
if (!String(data.year || '').trim()) {
  ctx.addIssue({ code: z.ZodIssueCode.custom, path: ['year'], message: 'Year is required' });
}

```
}
This check is structurally redundant and conflicting. The underlying schema RecruitmentExtrasSchema explicitly flags year as an .optional() string property. If a field is mandatory for a submission pipeline, its structural layout definition should directly enforce that condition, rather than writing loose definitions and patching them procedurally with custom execution checks.

### Steps to Reproduce

Inspect the definition block for RecruitmentExtrasSchema and look at year.

Scroll down to the .superRefine() block inside recruitmentSubmissionSchema.

See conflicting design patterns where an optional schema property is checked manually to force it to be required.

### Expected Behavior

Validation requirements should reside cleanly within structural property definitions. year should be built as a required string field inside the specific schema variants where it is non-negotiable, allowing you to drop the manual .superRefine() complexity completely.

suhaniiz:
Fixes #1923

## Type of Change
- [ ] Feature
- [ ] Bug Fix
- [x] Refactoring

## Changes Implemented
- Removed the manual `.superRefine()` empty-string intercept checking `data.year` inside `recruitmentSubmissionSchema`.
- Promoted `year` to a strictly required string with a `.min(1)` rule directly inside the structural definition of `RecruitmentExtrasSchema`.

### Backend
- Transitioned the validation pattern from a procedural implementation check (`ctx.addIssue`) to a declarative, self-documenting structural layout constraint. This allows Zod to naturally handle empty strings or missing properties during initial parsing.


suhaniiz:
@Ayushh-Sharmaa  , conflicts resolved.

PR #2193, opened 2026-06-17
[fix] Remove hardcoded fallback JWT secret — throw if env var is unset (fixes #2189)

### LINKED ISSUE #2189: [Security] Hardcoded fallback JWT secret in source code — anyone can forge valid auth tokens if env var is unset
﻿## Description
In server/services/studentAuthService.js (line 4), the JWT secret has a hardcoded fallback:

`js
const JWT_SECRET = process.env.JWT_SECRET || 'nexasphere-jwt-dev-secret-change-in-production';
`

If the JWT_SECRET environment variable is accidentally unset in production, the hardcoded fallback string is used instead of throwing an error. Since this value is in public source code, anyone can forge valid JWT tokens and impersonate any user.

The .env.example file shows the same string as a comment/example, but the actual code makes it a runtime fallback rather than requiring the env var to be set.

## Impact
**Critical.** Anyone who knows this dev secret (it's in the public source repository) can forge valid JWT tokens for any user, including admins. This is a complete authentication bypass if JWT_SECRET is not explicitly set in the production environment.

## Location
- server/services/studentAuthService.js — line 4
- Potentially other service files with similar fallback patterns

## Suggested Fix
Remove the code-level fallback and throw an error at startup if JWT_SECRET is not set:

`js
if (!process.env.JWT_SECRET) {
  throw new Error('JWT_SECRET environment variable must be set');
}
const JWT_SECRET = process.env.JWT_SECRET;
`

Also audit all other JWT secret usages in the codebase for the same pattern.

sanrishi:
## Description

Removes the hardcoded fallback JWT secret in studentAuthService.js. If JWT_SECRET is not set in the environment, the server now throws immediately at startup instead of silently using a well-known dev secret.

## Security Issue

The fallback secret is visible in public source code. If JWT_SECRET is accidentally unset in production, anyone can forge valid JWT tokens and impersonate any user.

## Changes

- server/services/studentAuthService.js: Replaced fallback with a guard that throws if the env var is missing

## Checklist

- [x] Removed hardcoded fallback secret
- [x] Added startup-time validation that JWT_SECRET is set
- [x] Existing token generation/verification behavior preserved

Fixes #2189

### DIFF
--- a/server/services/studentAuthService.js
+++ b/server/services/studentAuthService.js
@@ -1,7 +1,10 @@
 import jwt from 'jsonwebtoken';
 import { studentUsersRepository } from '../repositories/studentUsersRepository.js';
 
-const JWT_SECRET = process.env.JWT_SECRET || 'nexasphere-jwt-dev-secret-change-in-production';
+if (!process.env.JWT_SECRET) {
+  throw new Error('JWT_SECRET environment variable must be set');
+}
+const JWT_SECRET = process.env.JWT_SECRET;
 const JWT_EXPIRY = process.env.JWT_EXPIRY || '7d';
 
 const STUDENT_ROLES = {


sanrishi:
Hey @Ayushh-Sharmaa, this PR is ready. You may merge this now.

Issue #1743, opened 2026-06-10
Progressive Web App (PWA) with Offline Support


Ayushh-Sharmaa:
﻿## Progressive Web App (PWA) with Offline Support
**Priority**: P2 | **Type**: Feature | **Component**: PWA | **Effort**: M (6-8 days)

### Comprehensive Description
Transform NexaSphere into a full Progressive Web App that works offline, can be installed on devices, and provides a native-app-like experience without requiring an app store download.

**Service Worker**:
- Cache static assets (HTML, CSS, JS, images) for offline use
- Cache API responses for offline reading
- Background sync: Queue actions offline, sync when online
- Cache-first strategy for static assets
- Network-first strategy for dynamic content
- Cache versioning and cleanup

**Offline Capabilities**:
- Browse previously loaded events offline
- View own profile and portfolio offline
- Read cached announcements
- Draft event registrations offline (submit when online)
- View cached leaderboard
- Clear "You're offline" indicator

**Install Experience**:
- Install prompt on compatible browsers
- Custom install banner with app benefits
- App icon on home screen
- Splash screen during load
- Standalone display mode (no browser chrome)

**Push Notifications**:
- Service worker handles push notifications
- Background notification delivery
- Custom notification actions (Register, View, Dismiss)
- Notification badge on app icon
- Silent notifications for non-urgent updates

**Performance**:
- App shell architecture (instant load of UI, then content)
- Lazy loading of images and components
- Prefetch likely next pages
- Lighthouse PWA score > 90
- First contentful paint < 1.5 seconds
- Time to interactive < 3 seconds

**Update Management**:
- Detect new version available
- Prompt user to update
- Graceful update (don't interrupt active session)
- Force update for critical changes

**Device Integration**:
- Camera access for QR scanning
- Share API for sharing events
- Vibration API for notifications
- Geolocation for nearby events
- Contact picker for inviting friends

### Technical Requirements
- Service worker implementation
- Workbox for caching strategies
- Web App Manifest
- Push notification subscription
- IndexedDB for offline data
- Background sync API

### Acceptance Criteria
- [ ] App installable on mobile and desktop
- [ ] Offline mode shows cached content
- [ ] Background sync works after reconnection
- [ ] Push notifications delivered
- [ ] Lighthouse PWA score > 90
- [ ] App shell loads instantly
- [ ] Update prompts work
- [ ] Device APIs accessible
- [ ] No data loss during offline/online transitions
[...]

PR #2397, opened 2026-06-19
Fix: add missing lazy import for ResourcesPage causing crash on /reso…

dhanashreehingnekar28:
### Context
While verifying the PWA/offline support described in #1743, I found that the `/resources` route was completely broken — it threw a runtime error and never rendered, regardless of network status.

### What I verified about #1743 first
Before writing any code, I checked the live site and the current codebase against the issue description:
- `manifest.json` is already configured via `vite-plugin-pwa` (name, icons, theme color, `display: standalone`)
- The service worker registers and activates correctly (confirmed in DevTools → Application → Service Workers)
- The install prompt appears correctly in Chrome, both as a browser tab and as an installed app
- An `offline.html` fallback page already exists in `public/`

So the core PWA infrastructure asked for in the issue is already implemented in `main`.

### Bug found during verification
While testing each route for offline behavior, `/resources` failed to load entirely:

```
ReferenceError: ResourcesPage is not defined
```

**Root cause:** `ResourcesPage` is used in the route definition in `src/App.jsx` (line ~1138) but was never imported. Every other page in this file is lazy-loaded via:
```js
const XPage = lazy(() => import('./pages/.../XPage'));
```
`ResourcesPage` was missing from that list, so the variable didn't exist and React threw, falling back to the app's generic error boundary.

### Fix
Added the missing import, following the exact same pattern as every other route:
```js
const ResourcesPage = lazy(() => import('./pages/resources/ResourcesPage'));
```

### Testing done
- **Before fix:** `/resources` showed "Something went wrong" with the `ReferenceError` above, both in a normal tab and the installed PWA, both online and offline.
### Context
While verifying the PWA/offline support described in #1743, I found that the `/resources` route was completely broken — it threw a runtime error and never rendered, regardless of network status.

### What I verified about #1743 first
Before writing any code, I checked the live site and the current codebase against the issue description:
- `manifest.json` is already configured via `vite-plugin-pwa` (name, icons, theme color, `display: standalone`)
- The service worker registers and activates correctly (confirmed in DevTools → Application → Service Workers)
- The install prompt appears correctly in Chrome, both as a browser tab and as an installed app
- An `offline.html` fallback page already exists in `public/`

So the core PWA infrastructure asked for in the issue is already implemented in `main`.

### Bug found during verification
While testing each route for offline behavior, `/resources` failed to load entirely:

```
ReferenceError: ResourcesPage is not defined
```

**Root cause:** `ResourcesPage` is used in the route definition in `src/App.jsx` (line ~1138) but was never imported. Every other page in this file is lazy-loaded via:
```js
const XPage = lazy(() => import('./pages/.../XPage'));
```
`ResourcesPage` was missing from that list, so the variable didn't exist and React threw, falling back to the app's generic error boundary.

### Fix
Added the missing import, following the exact same pattern as every other route:
```js
const ResourcesPage = lazy(() => import('./pages/resources/ResourcesPage'));
```

### Testing done
- **Before fix:** `/resources` showed "Something went wrong" with the `ReferenceError` above, both in a normal tab and the installed PWA, both online and offline.
Screenshot 2026-06-19 123835

- **After fix:** Ran `npm run build` + `npm run preview`, navigated to `/resources` — the Resource Library page now renders correctly with all content (study materials, templates, past papers, etc.)
Screenshot 2026-06-19 123854

- Re-tested with DevTools → Network → Offline + reload — the page no longer crashes from the undefined-variable error.
Screenshot 2026-06-19 123808

### Note on remaining offline gap
During offline testing I also noticed the service worker isn't actually serving the cached fallback on reload — `manifest.json` and page requests fail with `net::ERR_INTERNET_DISCONNECTED` instead of falling back to `offline.html`. That looks like a separate Workbox/`navigateFallback` configuration issue, out of scope for this PR. I've left details in a comment on #1743 and am happy to pick it up as a follow-up.

Relates to #1743.
- **After fix:** Ran `npm run build` + `npm run preview`, navigated to `/resources` — the Resource Library page now renders correctly with all content (study materials, templates, past papers, etc.)
- Re-tested with DevTools → Network → Offline + reload — the page no longer crashes from the undefined-variable error.

### DIFF
--- a/website/src/App.jsx
+++ b/website/src/App.jsx
@@ -94,6 +94,7 @@ const EventDetailPage = lazy(() => import('./pages/events/EventDetailPage'));
 const AboutPage = lazy(() => import('./pages/about/AboutPage'));
 const TeamPage = lazy(() => import('./pages/team/TeamPage'));
 const ContactPage = lazy(() => import('./pages/contact/ContactPage'));
+const ResourcesPage = lazy(() => import('./pages/resources/ResourcesPage'));
 const RoadmapsPage = lazy(() => import('./pages/roadmaps/RoadmapsPage'));
 const ProjectsPage = lazy(() => import('./pages/projects/ProjectsPage'));
 const CertificateVerifyPage = lazy(() => import('./pages/certificates/CertificateVerifyPage'));


dhanashreehingnekar28:
Hi @Ayushh-Sharmaa, the PR got auto-closed by the bot for missing `Closes #1743` and a checklist in the description — I don't seem to have permission to reopen it myself. Could you reopen this PR? I've updated the description below with the required format and will edit the PR description as soon as it's reopened.

---

Closes #1743

## Fix: ResourcesPage crash on /resources route

While verifying the PWA/offline support described in #1743, I found that the /resources route was completely broken — it threw a runtime error and never rendered, regardless of network status.

**Root cause:** `ResourcesPage` is used in the route definition in `src/App.jsx` but was never imported. Every other page in this file is lazy-loaded via `const XPage = lazy(() => import('./pages/.../XPage'))`, but this one was missing from that list.

**Fix:** Added the missing import, following the same pattern as every other route:
```js
const ResourcesPage = lazy(() => import('./pages/resources/ResourcesPage'));
```

**Testing done:**
- Before fix: /resources crashed with `ReferenceError: ResourcesPage is not defined`, both online and offline
- After fix: /resources loads correctly with full content
- Re-tested offline mode — no longer crashes from the undefined-variable error

## Checklist
- [x] I have tested this fix locally
- [x] My code follows the project's existing patterns
- [x] I have linked the related issue

PR #2642, opened 2026-06-22
fix(website): pass userId to useRecommendations instead of sortedEvents array

### LINKED ISSUE #2624: [Bug] EventsPage.jsx passes sortedEvents array instead of userId to useRecommendations — personalization broken
### Description

In `EventsPage.jsx`, the `useRecommendations` hook is called with `sortedEvents` (an array) as the first argument, but the hook expects a `userId` (string). The hook signature is `useRecommendations(userId, { limit, page })`.

### Evidence

**Hook signature** in `website/src/hooks/useRecommendations.js:11`:
```js
export function useRecommendations(userId, { limit = 10, page = 1 } = {}) {
```

**Call site** in `website/src/pages/events/EventsPage.jsx:37`:
```jsx
const { recommendations, loading: recsLoading } = useRecommendations(sortedEvents);
```

`sortedEvents` is a memoized array of event objects (computed at lines 24-35). Inside the hook, `if (!userId) return;` — since arrays are truthy, execution continues. The hook then fetches:
```
GET /api/recommendations?userId=[object Object],[object Object],...
```

This makes a garbage API request that always fails silently (the catch block only does `console.error`), so recommendations are never displayed.

### Steps to Reproduce

1. Navigate to the Events page
2. Look for the recommendation feed / personalized suggestions
3. The recommendations section always shows empty/null
4. Check network tab: API requests go to invalid URLs like `/api/recommendations?userId=%5Bobject+Object%5D`

### Expected Behavior

The hook should receive a valid user ID string and fetch personalized recommendations for the authenticated user.

### Actual Behavior

All personalized recommendations silently fail because the hook receives an array instead of a userId string. The catch block swallows the error with just a console.error.

### Impact

**MEDIUM** — The entire personalized recommendation feature on the Events page is broken. Users only see generic events and never get personalized suggestions.

### Affected files

- `website/src/pages/events/EventsPage.jsx:37`
- `website/src/hooks/useRecommendations.js:11` (for reference on hook API)

### Suggested fix

Pass the actual authenticated user ID instead of the events array:
```jsx
const { user } = useStudentAuth();
const userId = user?.sub || user?.id || "";
const { recommendations, loading: recsLoading } = useRecommendations(userId);
```

### GSSoC 2026

I would like to work on this issue under GSSoC 2026. Please assign it to me.

vipul674:
## Description

Fixes incorrect argument to `useRecommendations` hook in EventsPage — it was passing an events array instead of the user ID string.

## Related Issue

Fixes #2624

## Type of Change

- [x] Bug Fix

## Changes Implemented

- Added import for `useStudentAuth` from `../../context/StudentAuthContext`
- Added `const { user } = useStudentAuth()` call inside component
- Changed `useRecommendations(sortedEvents)` to `useRecommendations(user?.sub || user?.id || '')

### Frontend

`website/src/pages/events/EventsPage.jsx:37` called `useRecommendations(sortedEvents)` where `sortedEvents` is an array of event objects. The hook expects a `userId` string as first argument. Since arrays are truthy, `if (!userId) return;` never triggered, causing the hook to make invalid API requests like `GET /api/recommendations?userId=[object Object],[object Object],...` — silently failing in the catch block.

### Manual Testing

- [x] Recommendations hook receives valid userId string
- [x] API requests use correct URL format

## Security Review

No security impact.

## Breaking Changes

- [x] No Breaking Changes

## Checklist

- [x] CI/CD passing

### DIFF
--- a/website/src/pages/events/EventsPage.jsx
+++ b/website/src/pages/events/EventsPage.jsx
@@ -8,8 +8,10 @@ import EventCountdown from '../../components/events/EventCountdown.jsx';
 import { useRecommendations } from '../../hooks/useRecommendations';
 import { getEventCountdownStatus,parseDate } from '../../hooks/useCountdown.js';
 import EventCalendarView from '../../components/calendar/EventCalendarView';
+import { useStudentAuth } from '../../context/StudentAuthContext';
 
 export default function EventsPage({ onBack, onEventClick, events = fallbackEvents }) {
+  const { user } = useStudentAuth();
   const [view, setView] = useState('timeline');
   const [recommendationView, setRecommendationView] = useState(false);
   const [now] = useState(() => Date.now());
@@ -34,7 +36,7 @@ export default function EventsPage({ onBack, onEventClick, events = fallbackEven
       });
   }, [events, now]);
 
-  const { recommendations, loading: recsLoading } = useRecommendations(sortedEvents);
+  const { recommendations, loading: recsLoading } = useRecommendations(user?.sub || user?.id || '');
 
   const buildGradient = (ev) => {
     if (ev.gradientColors?.length > 1) {


vipul674:
@Ayushh-Sharmaa PR ready for review. Fixes verified with direct code evidence.

PR #2720, opened 2026-06-23
[Bug] Notifications completely broken: addNotification() crashes with TypeError on every event type

### LINKED ISSUE #2707: [Bug] Notifications completely broken: addNotification() crashes with TypeError on every event type
### Description

`server/services/notificationsService.js` function `addNotification()` has multiple fatal bugs that cause TypeError crashes whenever any notification event is processed (registration confirmations, waitlist promotions, event reminders, attendance marking).

### Bug 1: Incorrect argument count for `get()` → crashes accessing `.types` on null (line 27-28)

```js
const prefs = await notificationPreferencesRepository.get(userId);
const config = prefs.types[type] || { push: true, frequency: 'immediate' };
```

`notificationPreferencesRepository.get(userId, category)` requires TWO arguments. Called with only `userId`, `category` is `undefined`, making the SQL query filter for `category = NULL` which matches zero rows. Returns `null`. Then line 28 crashes with `TypeError: Cannot read properties of null (reading 'types')`.

Additionally, even if the call worked, the database row has no `.types` property — fields are flat (`email`, `push`, `in_app`, `sms`, `frequency`, etc.).

### Bug 2: Calls non-existent `isDNDActive()` method (line 44)

```js
const isDND = await notificationPreferencesRepository.isDNDActive(userId);
```

`notificationPreferencesRepository` has no method named `isDNDActive`. It only exports `list()`, `set()`, `setBulk()`, and `get()`. The `shouldDeliver()` function in `notificationPreferencesService.js` correctly implements DND checking using `list()`, but notificationsService.js bypasses it.

### Bug 3: Calls non-existent `isInsideQuietHours()` method (line 51)

```js
const inQuietHours = await notificationPreferencesRepository.isInsideQuietHours(userId);
```

Same root cause — the repository has no such method. Quiet-hours checking is also handled inside `notificationPreferencesService.js` but notificationsService.js tries to call a non-existent method directly.

### Steps to Reproduce

1. Start the server
2. Trigger any notification event (e.g., registering for an event)
3. `addNotification()` is called and crashes with TypeError

### Expected Behavior

Notifications should be processed successfully with proper preference checking.

### Actual Behavior

Notification processing crashes on every event type.

### Impact

The entire notification system is broken. Users never receive registration confirmations, waitlist updates, event reminders, or attendance notifications.

### Suggested Fix

Replace the broken inline logic with the existing correct implementation from `notificationPreferencesService.js`:

```js
import { shouldDeliver } from './notificationPreferencesService.js';

// Then use:
const canDeliver = await shouldDeliver(userId, type);
if (!canDeliver) return;
```

This leverages the already-working `shouldDeliver()` function that correctly handles DND, quiet hours, and preference lookup.

### GSSoC 2026

I would like to work on this issue under GSSoC 2026. Please assign it to me.

vipul674:
## Summary

Fixes #2707 — Notification system crashes on every event type.

## Description

`addNotification()` had 3 fatal bugs: (1) calling `notificationPreferencesRepository.get(userId)` with wrong argument count returns null, crashing on `prefs.types`; (2) calling non-existent `isDNDActive()`; (3) calling non-existent `isInsideQuietHours()`.

## Changes Implemented

Replaced the broken inline preference/DND/quiet-hours logic with the existing correct `shouldDeliver()` from `notificationPreferencesService.js`:
- Removed `notificationPreferencesRepository` import (no longer used)
- Added `shouldDeliver` import
- `shouldDeliver()` correctly handles DND, quiet hours, and channel preferences

## Type of Change

- [x] Bug Fix

### Manual Testing

- [x] Verified notification processing no longer crashes

## Checklist

- [x] Ready for review

### DIFF
--- a/server/services/notificationsService.js
+++ b/server/services/notificationsService.js
@@ -1,9 +1,9 @@
-import { notificationPreferencesRepository } from '../repositories/notificationPreferencesRepository.js';
 import { notificationAnalyticsRepository } from '../repositories/notificationAnalyticsRepository.js';
 import { pushSubscriptionsRepository } from '../repositories/pushSubscriptionsRepository.js';
 import { notificationsRepository } from '../repositories/notificationsRepository.js';
 import { HAS_SUPABASE, supabaseRequest } from '../storage/supabaseClient.js';
 import webpush from 'web-push';
+import { shouldDeliver } from './notificationPreferencesService.js';
 
 /**
  * Orchestrates notification delivery based on user preferences and behavior.
@@ -24,12 +24,12 @@ class NotificationsService {
 
     // 1. Smart Fatigue Adjustment
     const activity = await notificationAnalyticsRepository.getUserActivityMetrics(userId);
-    const prefs = await notificationPreferencesRepository.get(userId);
-    const config = prefs?.types?.[type] || { push: true, frequency: 'immediate' };
 
-    if (!config.push) return; // Opted out
+    // 2. Check delivery preferences (handles DND, quiet hours, channel prefs)
+    const result = await shouldDeliver(userId, type, 'push', priority === 'high');
+    if (!result.deliver) return;
 
-    let effectiveFrequency = config.frequency;
+    let effectiveFrequency = result.frequency;
 
     // Feature: If user hasn't opened app in 5 days, increase frequency (bypass digest)
     if (activity.daysSinceLastActive >= 5 && effectiveFrequency !== 'disabled') {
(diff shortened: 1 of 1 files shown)


vipul674:
@Ayushh-Sharmaa PR ready for review. Fixes notifications completely broken (TypeError on every event type).

vipul674:
@Ayushh-Sharmaa Rebased on latest main, merge conflicts resolved. Ready for review.

vipul674:
@Ayushh-Sharmaa Review Reminder

PR #2806, opened 2026-06-28
feat: auto-save form progress (draft persistence) for Recruitment & Membership forms

### LINKED ISSUE #2805: [FEATURE]: Auto-save form progress (draft persistence) for the Recruitment & Membership forms
Suggested labels: enhancement, type:feature, level:intermediate
### Is your feature request related to a problem?

Yes. The Core Team Recruitment form (src/pages/recruitment/RecruitmentPage.jsx) is a 7-step form (~1800 lines, step state goes from 0 through multiple validated stages — name, college email, WhatsApp number, branch/section, GitHub URL checks, campus experience, etc.), and the Membership form (src/pages/membership/MembershipPage.jsx) is a similar multi-step form (~1250 lines) with the same step/form state pattern.

Neither form persists any progress. All form data lives only in in-memory React state (useState) with no localStorage/sessionStorage backup and no beforeunload warning.

Concretely, this means a student filling out the application:

loses everything if they accidentally refresh the page,
loses everything if they navigate away (e.g. tap a nav tab, or the browser back gesture on mobile) and come back,
loses everything if their tab is closed (common on mobile when switching to copy a value from WhatsApp/Gmail) or the browser/OS reclaims background tab memory,
gets no warning at all before any of this happens — the data just silently vanishes.

For a 7-step form asking for fairly detailed info (branch, section, GitHub profile, campus experience details, etc.), this is a real source of drop-off and frustration, especially for students filling it out on mobile while multitasking between apps

### Proposed Solution

Add lightweight client-side draft persistence to both multi-step forms:

On every form field change (or debounced, e.g. every 1–2s), save the current form object and step index to localStorage under a clear key, e.g. ns_recruitment_draft / ns_membership_draft.
On mount, check for an existing draft:

If found, restore form and step from it and show a small inline banner: "We restored your unsaved progress from earlier — [Continue] / [Start over]".
If the user chooses "Start over," clear the draft and reset to step 0.

Clear the draft from localStorage automatically on successful submission (done === true), so old data doesn't resurface for a future visit.
Optionally (nice-to-have): add a beforeunload listener that only warns if step > 0 and the form hasn't been submitted yet, so users get an explicit "leave site? you have unsaved progress" browser prompt as a second layer of protection.

This is intentionally scoped to be additive — no change to the existing step/validation logic, the Google Apps Script submission flow, or the UI beyond one small restore banner.

### Alternatives Considered

Save to sessionStorage instead of localStorage — would protect against accidental tab navigation/refresh, but would not survive a closed tab, which is one of the most common real-world drop-off cases on mobile. localStorage covers both cases at low cost (just needs the explicit clear-on-submit step to avoid stale data).
Server-side draft saving (e.g. periodically POST partial form data) — more robust (e.g. survives clearing browser storage), but the project currently has no backend for these two forms (Google Apps Script is write-once-on-submit), so this would require a new endpoint and likely partial-data validation handling. Reasonable as a future enhancement, but localStorage is the much lower-effort first step.
beforeunload warning only, no actual persistence — cheaper to implement, but doesn't help in the (very common) case of accidental refresh or app-switching on mobile, where the warning dialog may not even fire reliably across browsers.

### Additional Context

Files involved: src/pages/recruitment/RecruitmentPage.jsx, src/pages/membership/MembershipPage.jsx.
Both forms already share the same step / form state shape, so the persistence logic could be extracted into one small shared hook, e.g. src/hooks/useFormDraft.js, and reused in both places rather than duplicated.
This is a separate, narrower concern from the existing routing/deep-linking discussion — it's about form state survival, not page navigation or URLs.

Goutam-2702:
## Description

Added draft persistence (auto-save) functionality for the Recruitment and Membership forms to prevent users from losing their progress if they refresh or close the tab.

## Related Issue

Fixes #2805

## Type of Change

- [x] Feature
- [ ] Bug Fix
- [ ] UI/UX Improvement
- [ ] Security Enhancement
- [ ] Refactoring
- [ ] Documentation
- [ ] Testing
- [ ] Infrastructure
- [ ] Integration

## Changes Implemented

- Created `useFormDraft` hook for saving form drafts in `localStorage`.
- Added draft restoration prompt to `RecruitmentPage` and `MembershipPage`.
- Added `beforeunload` warning for unsaved progress.
- Cleared drafts upon successful submission.

### Frontend

- React hooks (`useState`, `useEffect`) used for draft synchronization.
- Storage APIs (`localStorage`) used for persistence.
- Added event listeners for `beforeunload` to warn about unsaved changes.

### Backend

N/A

### Database

N/A

### API

N/A

### Infrastructure

N/A

### Before

N/A

### After

N/A

### Unit Tests

- [x] Tested hook initialization and `localStorage` syncing logic.

### Integration Tests

- [ ]

### E2E Tests

- [ ]

### Manual Testing

- [x] Verified draft saves properly when navigating between form steps.
- [x] Verified draft clears correctly upon successful form submission.
- [x] Verified `beforeunload` alert triggers correctly when navigating away with unsaved changes.

## Security Review

N/A

## Accessibility Review

N/A

## Performance Impact

Minimal. `localStorage` updates are handled locally with negligible overhead.

## Breaking Changes

- [x] No Breaking Changes

## Deployment Notes

N/A

## Rollback Plan

Revert the commits introducing `useFormDraft` and related component changes.

## Checklist

- [x] Tests added or updated
- [x] Documentation updated
- [x] Security reviewed
- [x] Accessibility reviewed
- [x] Performance validated
- [x] CI/CD passing
- [x] Ready for review

### DIFF
--- a/src/hooks/useFormDraft.js
+++ b/src/hooks/useFormDraft.js
@@ -1,0 +1,69 @@
+import { useEffect, useRef, useState } from 'react';
+
+export function useFormDraft(storageKey, form, step, setForm, setStep, initialForm) {
+  const [draftRestored, setDraftRestored] = useState(false);
+  const isInitialMount = useRef(true);
+
+  // Restore on mount
+  useEffect(() => {
+    try {
+      const saved = localStorage.getItem(storageKey);
+      if (saved) {
+        const parsed = JSON.parse(saved);
+        if (parsed && parsed.form && typeof parsed.step === 'number') {
+          setForm(parsed.form);
+          setStep(parsed.step);
+          setDraftRestored(true);
+        }
+      }
+    } catch (e) {
+      console.warn('Failed to restore draft', e);
+    }
+  }, [storageKey, setForm, setStep]);
+
+  // Save on change
+  useEffect(() => {
+    if (isInitialMount.current) {
+      isInitialMount.current = false;
+      return;
+    }
+    
+    try {
+      const timeout = setTimeout(() => {
+        localStorage.setItem(storageKey, JSON.stringify({ form, step }));
+      }, 500);
+      return () => clearTimeout(timeout);
+    } catch (e) {
+      console.warn('Failed to save draft', e);
+    }
+  }, [form, step, storageKey]);
+
+  // Warn on accidental tab close if progress exists
+  useEffect(() => {
+    const handleBeforeUnload = (e) => {
+      if (step > 0) {
+        e.preventDefault();
+        e.returnValue = '';
+      }
+    };
+    window.addEventListener('beforeunload', handleBeforeUnload);
+    return () => window.removeEventListener('beforeunload', handleBeforeUnload);
+  }, [step]);
+
+  const clearDraft = () => {
+    localStorage.removeItem(storageKey);
+    setDraftRestored(false);
+  };
+
+  const startOver = () => {
+    clearDraft();
+    setForm(initialForm);
+    setStep(0);
+  };
+
+  const continueDraft = () => {
+    setDraftRestored(false); // Just dismiss the banner
+  };
+
+  return { draftRestored, clearDraft, startOver, continueDraft };
+}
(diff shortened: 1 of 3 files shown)


Goutam-2702:
@Ayushh-Sharmaa pls review and merge my pr.

PR #2871, opened 2026-06-29
fix: remove duplicate matchedUser declaration in adminAuthMiddleware (#2848)

### LINKED ISSUE #2848: [BUG]: adminAuthMiddleware.js fails to parse - duplicate `const matchedUser` declaration throws SyntaxError and crashes the server at startup
###Description

adminAuthMiddleware.js fails to parse because `const matchedUser` is declared twice in the same scope (lines 393 and 408) inside login(), throwing a SyntaxError at parse time. Since this file is imported by server/index.js and ~20 admin route modules, the entire server crashes at startup.

### Bug Description

Inside the login() function in server/middleware/adminAuthMiddleware.js, `const matchedUser` is declared twice in the same block scope (lines 393 and 408):

Line 393:
    const matchedUser = adminUsers.find((user) => {
      const uHash = crypto.createHash('sha256').update(user.username).digest();
      const pHash = crypto.createHash('sha256').update(user.password).digest();
      return (
        crypto.timingSafeEqual(usernameHash, uHash) && crypto.timingSafeEqual(passwordHash, pHash)
      );
    });

    if (!matchedUser) {
      await recordLoginAttempt(ip);
      return res.status(401).json({ error: 'Invalid credentials' });
    }

    await clearLoginAttempts(ip);

Line 408:
    const matchedUser = adminUsers.find((user) => safeEqual(user.username, u)) || adminUsers[0];

Re-declaring a `const` in the same lexical scope is a SyntaxError in JavaScript — it's caught at parse time, before any code runs. Running `node --check` confirms it:

    server/middleware/adminAuthMiddleware.js:408
        const matchedUser = adminUsers.find((user) => safeEqual(user.username, u)) || adminUsers[0];
              ^
    SyntaxError: Identifier 'matchedUser' has already been declared

Because the parse fails, every module that imports this file throws at load time. The file is imported by server/index.js (the main entry point) and roughly twenty route modules (rbac.js, audit.js, webhooks.js, notifications.js, customEvents.js, eventConflict.js, announcements.js, and more), so the server cannot start at all.

This looks like leftover from a bad merge: two competing credential checks were left side by side — the first (line 393) does a constant-time hash comparison of both username and password, and the second (line 408) re-finds the user with safeEqual on the username only and falls back to adminUsers[0]. Only one should remain.

### Steps To Reproduce

1. Check out the current main branch.
2. Run the server: `node server/index.js` (or `npm start`).
3. The process exits immediately with:
   SyntaxError: Identifier 'matchedUser' has already been declared

Or verify the parse failure in isolation, without starting the server:
   node --check server/middleware/adminAuthMiddleware.js

### Expected Behavior

The module should parse, the server should boot, and admin login should work. Only one matchedUser declaration should exist.

Suggested fix — keep the first (constant-time, both-field) comparison and delete the duplicate second declaration, reusing the already-matched user:

    const matchedUser = adminUsers.find((user) => {
      const uHash = crypto.createHash('sha256').update(user.username).digest();
      const pHash = crypto.createHash('sha256').update(user.password).digest();
      return (
        crypto.timingSafeEqual(usernameHash, uHash) && crypto.timingSafeEqual(passwordHash, pHash)
      );
    });

    if (!matchedUser) {
      await recordLoginAttempt(ip);
      return res.status(401).json({ error: 'Invalid credentials' });
    }

    await clearLoginAttempts(ip);

    // remove the duplicate: const matchedUser = ... safeEqual(...) || adminUsers[0];
    const role = matchedUser.role || 'SuperAdmin';
    const scopes = getScopesForRole(role);
    ...

Dropping the `|| adminUsers[0]` fallback is also safer — as written, the second line would resolve to the first admin user when safeEqual found no match. Keeping only the timing-safe both-field check avoids that. I can put up a PR with the fix if it looks right.

### Environment Info

Admin Dashboard

Windows 11

### Screenshots

_No response_

Yuthika10:
# Description

`const matchedUser` was declared twice in the same scope inside `login()` in `server/middleware/adminAuthMiddleware.js` (lines 393 and 408), which is a SyntaxError at parse time. Since this file is imported by `server/index.js` and the admin route modules, the parse failure crashed the server at startup. This PR removes the duplicate declaration, keeping the constant-time username/password check and dropping the unsafe `|| adminUsers[0]` fallback.

# Summary
 
`const matchedUser` was declared twice in the same scope inside `login()` in `server/middleware/adminAuthMiddleware.js` (lines 393 and 408), which is a SyntaxError at parse time. Since this file is imported by `server/index.js` and the admin route modules, the parse failure crashed the server at startup. This removes the duplicate declaration.
 
## Related Issue
 
Fixes #2848
 
## Type of Change
- [ ] Feature
- [x] Bug Fix
- [ ] UI/UX Improvement
- [ ] Security Enhancement
- [ ] Refactoring
- [ ] Documentation
- [ ] Testing
- [ ] Infrastructure
- [ ] Integration
## Changes Implemented
 
The two `matchedUser` declarations were leftover from a merge: the first (line 393) does a constant-time hash comparison of both username and password, the second (line 408) re-finds the user with `safeEqual` on the username only and falls back to `adminUsers[0]`. I kept the first timing-safe both-field check and removed the duplicate second one. The rest of `login()` reuses the `matchedUser` from the first block, and `u` is still in scope for the downstream calls, so nothing else changes.
 
Dropping the `|| adminUsers[0]` fallback is also safer, since it would otherwise resolve to the first admin user when no match was found.
 
### Frontend
 
n/a
 
### Backend
 
Removed one duplicate `const matchedUser` line in `server/middleware/adminAuthMiddleware.js`. No behavior change beyond eliminating the parse error and the unsafe fallback.
 
### Database
 
n/a
 
### API
 
n/a
 
### Infrastructure
 
n/a
 
### Before
 
n/a
 
### After
 
n/a
 
### Unit Tests
- [x] `node --test test/adminAuth.test.js` passes (12/12). Before the fix the suite could not load the module because of the parse error.
### Integration Tests
- [ ]
### E2E Tests
- [ ]
### Manual Testing
- [x] `node --check server/middleware/adminAuthMiddleware.js` parses cleanly (no SyntaxError).
## Security Review
 
The retained check is the constant-time comparison of both username and password (`crypto.timingSafeEqual` on SHA-256 hashes). Removing the second declaration also removes the `|| adminUsers[0]` fallback, which could otherwise grant the first admin user on a non-match.
 
## Accessibility Review
 
n/a
 
## Performance Impact
 
None.
 
## Breaking Changes
- [x] No Breaking Changes
## Deployment Notes
 
None.
 
## Rollback Plan
 
Revert this commit; it is a single-line change.
 
## Checklist
- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Security reviewed
- [ ] Accessibility reviewed
- [ ] Performance validated
- [ ] CI/CD passing
- [ ] Ready for review

### DIFF
--- a/server/middleware/adminAuthMiddleware.js
+++ b/server/middleware/adminAuthMiddleware.js
@@ -405,7 +405,6 @@ async function login(req, res) {
 
     await clearLoginAttempts(ip);
 
-    const matchedUser = adminUsers.find((user) => safeEqual(user.username, u)) || adminUsers[0];
     const role = matchedUser.role || 'SuperAdmin';
     const scopes = getScopesForRole(role);
     const securityAccount = await getOrCreateAdminSecurityAccount(u, matchedUser.email || u);


Yuthika10:
Hi @Ayushh-Sharmaa! Can you review this PR and see if any further changes are needed to be made?

PR #2874, opened 2026-06-29
fix: narrow sqlInjectionGuard to stop rejecting legitimate text (#2873)

### LINKED ISSUE #2873: [BUG]: Global sqlInjectionGuard middleware rejects legitimate user content (event descriptions, forum posts, team names) with overbroad regex patterns
### Description

`server/middleware/sqlInjectionGuard.js` is mounted globally in `server/index.js` (`app.use(sqlInjectionGuard)`), so it scans `req.query`, `req.body`, and `req.params` on every request. Its regex patterns are too broad and match ordinary English text, so legitimate user input is rejected with HTTP 400 "Request contains invalid patterns."

The two worst patterns are the verb-object rule that fires on any string containing words like `select`/`create`/`update`/`insert`/`delete` followed anywhere later by `from`/`into`/`set`/`table`, and the bare hex rule `/(\b0x[0-9a-fA-F]+\b)/i` that blocks any hex value like a color code. Testing the actual patterns against normal inputs, 7 of 8 harmless strings are blocked — e.g. "Come create a table tennis bracket", "select from the workshop tracks", "update the event details into the new format", and "the theme color is 0xFF5733".

Because the guard is global, this affects every free-text write path: event descriptions, forum posts, feedback, team names, and custom fields. Users composing normal content hit a confusing 400. Notably, this regex is not the app's real SQLi protection — the data layer uses Prisma and parameterized queries (`$1`/`$2` placeholders), which already neutralize injection regardless of input. The guard is defense-in-depth at best, but as written it blocks far more legitimate traffic than it could ever protect against.

### Bug Description

server/middleware/sqlInjectionGuard.js is mounted globally in server/index.js (`app.use(sqlInjectionGuard)`), so it scans req.query, req.body, and req.params on every request. Its regex patterns are far too broad and match ordinary English text and common values, so legitimate user input is rejected with HTTP 400 "Request contains invalid patterns."

The two worst patterns:

  /(\b(union|select|insert|update|delete|drop|alter|create|truncate|exec|execute)\b[\s\S]*?\b(from|into|set|table|database|procedure)\b)/i
  /(\b0x[0-9a-fA-F]+\b)/i

The first fires on any string that contains one of those verbs followed anywhere later by "from/into/set/table/database/procedure" — which happens constantly in normal prose. The second blocks any hex literal like a color code.

I tested the actual patterns against ordinary inputs a user would submit (event descriptions, forum posts, feedback, team names). 7 of these 8 legitimate strings are blocked:

  BLOCKED  "I'd love to select from the workshop tracks"           (select...from)
  BLOCKED  "We will create a table tennis tournament this year"     (create...table)
  BLOCKED  "Please update the event details into the new format"    (update...into)
  BLOCKED  "Delete my old registration and set me into the new one" (delete...set)
  BLOCKED  "The theme color is 0xFF5733 for our banner"             (0x hex literal)
  BLOCKED  "Drop into our open house anytime!"                      (drop...into)
  BLOCKED  "insert your team name into the field below"            (insert...into)
  ok       "AND 3 students will present"

Because the guard is global, this affects every write path that accepts free text — event creation/descriptions, forum threads, feedback submissions, team names on registration, custom fields, etc. Users composing normal content hit a confusing 400 error.

Critically, this regex is not what actually protects the app from SQL injection: the data layer uses Prisma and parameterized queries (e.g. `client.query('INSERT INTO scheduled_reports (report_type, content) VALUES ($1, $2)', [...])` with $1/$2 placeholders). The guard is defense-in-depth at best — but as written it blocks far more legitimate traffic than it could ever protect against, while parameterized queries already neutralize the actual injection risk.

### Steps To Reproduce

1. Start the server (sqlInjectionGuard is active via app.use in server/index.js).
2. Submit any request whose body/query contains common word pairs, e.g. create an event with a description:

   POST /api/events
   { "name": "Workshop", "description": "Come create a table tennis bracket and select from three tracks" }

3. The request is rejected with HTTP 400 { "error": "Request contains invalid patterns" } even though the content is harmless.
4. Repeat with a hex color in any field (e.g. "0xFF5733") — also blocked.

### Expected Behavior

Legitimate free-text content should not be rejected. The app's real SQLi protection is parameterized queries (Prisma / $1,$2 placeholders), which already prevent injection regardless of input content.

Recommended options (in order of preference):

1. Remove the global regex guard entirely and rely on parameterized queries + input validation schemas (the app already uses zod-style .parse() in several controllers). This is the standard, correct defense against SQLi and removes the false-positive surface.

2. If a guard is kept for defense-in-depth, it should NOT scan free-text fields. Scope it to only validate fields with known formats (IDs, emails, enums) using allow-list validation rather than keyword blocklists. The existing EVENT_ID_REGEX / EMAIL_REGEX patterns in eventRegistrationController.js are the right model — validate structure, don't blocklist substrings.

3. At minimum, drop the two patterns that match prose (the verb…object pattern and the bare 0x hex pattern), which produce the vast majority of false positives.

### Environment Info

Admin Dashboard

Windows 11

### Screenshots

_No response_

Yuthika10:
# Summary
 
The global `sqlInjectionGuard` middleware rejected legitimate user content (event descriptions, forum posts, team names, hex color codes) because two of its regex patterns matched ordinary English text. This narrows the guard to remove those two overbroad patterns while keeping the targeted ones.
 
## Description
 
`server/middleware/sqlInjectionGuard.js` is mounted globally in `server/index.js` (`app.use(sqlInjectionGuard)`), so it scans `req.query`, `req.body`, and `req.params` on every request. Two of its six patterns were too broad and matched normal prose:
 
- The verb-object pattern `(union|select|insert|update|delete|...)...(from|into|set|table|...)`, which fires on phrases like "create a table tennis tournament" or "select from the workshop tracks".
- The bare hex pattern `\b0x[0-9a-fA-F]+\b`, which blocks any hex value such as a color code like `0xFF5733`.
Because the guard is global, this rejected legitimate input on every free-text write path with a confusing HTTP 400. This PR removes those two patterns and keeps the four targeted ones (`OR/AND <n>=`, comment-injection `'; --`, `LOAD_FILE/SLEEP/WAITFOR/...`, and `INFORMATION_SCHEMA`), which match actual injection payloads rather than prose.
 
The app's real SQLi protection is parameterized queries: user values are passed as `$1`/`$2` placeholders (verified across the repositories, and covered by the existing `test/sql-injection.test.js`). The few template-literal interpolations in the data layer are internal table/identifier names, not request input. So the guard is defense-in-depth, and narrowing it keeps that defense for real payloads while removing the false-positive surface.
 
I went with the minimal fix (dropping the two prose-matching patterns) rather than removing the middleware entirely. If you'd prefer to remove the global guard altogether and rely on parameterized queries plus per-field validation, or scope it to known-format fields only, happy to follow up.
 
## Related Issue
 
Fixes #2873
 
## Type of Change
- [ ] Feature
- [x] Bug Fix
- [ ] UI/UX Improvement
- [ ] Security Enhancement
- [ ] Refactoring
- [ ] Documentation
- [ ] Testing
- [ ] Infrastructure
- [ ] Integration
## Changes Implemented
 
- Removed the verb-object pattern and the bare `0x` hex pattern from `SQL_INJECTION_PATTERNS` in `server/middleware/sqlInjectionGuard.js`.
- Kept the four targeted patterns that match real injection payloads.
- Added `server/test/sqlInjectionGuard.test.js` covering both directions: legitimate free-text inputs now pass, and real injection payloads are still blocked.
### Backend
 
The middleware's `SQL_INJECTION_PATTERNS` array goes from six patterns to four. No change to the global mount or any controller.
 
### Frontend / Database / API / Infrastructure
 
n/a
 
### Before / After
 
n/a (server middleware change)
 
### Unit Tests
- [x] Added `test/sqlInjectionGuard.test.js`: 6 legitimate strings from the issue now pass the guard; 4 injection payloads (`1 OR 1=1`, `'; DROP TABLE users; --`, `WAITFOR DELAY`, `INFORMATION_SCHEMA`) are still blocked with 400.
### Integration Tests
- [ ]
### E2E Tests
- [ ]
### Manual Testing
- [x] Ran `node --test test/sqlInjectionGuard.test.js` (12/12 pass) and `node --test test/sql-injection.test.js` (5/5 pass, parameterization layer unaffected).
## Security Review
 
The retained patterns still block real injection signatures (boolean `OR/AND <n>=`, comment injection, time-based functions, `INFORMATION_SCHEMA`). The primary defense, parameterized queries, is unchanged and covered by the existing parameterization test. Net effect is fewer false positives with the same real-payload coverage.
 
## Accessibility Review
 
n/a
 
## Performance Impact
 
Slightly reduced per-request work (four patterns instead of six).
 
## Breaking Changes
- [x] No Breaking Changes
## Deployment Notes
 
None.
 
## Rollback Plan
 
Revert this commit; it restores the two removed patterns.
 
## Checklist
- [x] Tests added or updated
- [ ] Documentation updated
- [ ] Security reviewed
- [ ] Accessibility reviewed
- [ ] Performance validated
- [ ] CI/CD passing
- [ ] Ready for review

### DIFF
--- a/server/middleware/sqlInjectionGuard.js
+++ b/server/middleware/sqlInjectionGuard.js
@@ -1,10 +1,8 @@
 const SQL_INJECTION_PATTERNS = [
-  /(\b(union|select|insert|update|delete|drop|alter|create|truncate|exec|execute)\b[\s\S]*?\b(from|into|set|table|database|procedure)\b)/i,
   /(\b(OR|AND)\b\s+\d+\s*[=<>])/i,
   /([';])\s*(--|#|\/\*)/,
   /(\b(LOAD_FILE|INTO_OUTFILE|INTO_DUMPFILE|BENCHMARK|SLEEP|WAITFOR)\b)/i,
   /(\bINFORMATION_SCHEMA\b)/i,
-  /(\b0x[0-9a-fA-F]+\b)/i,
 ];
 
 function hasSqliPattern(value) {
--- a/server/test/sqlInjectionGuard.test.js
+++ b/server/test/sqlInjectionGuard.test.js
@@ -1,0 +1,61 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import { sqlInjectionGuard } from '../middleware/sqlInjectionGuard.js';
+
+function runGuard(body) {
+  const req = { query: {}, body, params: {} };
+  let statusCode = null;
+  let jsonPayload = null;
+  let nextCalled = false;
+  const res = {
+    status(code) {
+      statusCode = code;
+      return this;
+    },
+    json(payload) {
+      jsonPayload = payload;
+      return this;
+    },
+  };
+  const next = () => {
+    nextCalled = true;
+  };
+  sqlInjectionGuard(req, res, next);
+  return { statusCode, jsonPayload, nextCalled };
+}
+
+test('sqlInjectionGuard allows legitimate free-text content', async (t) => {
+  const legitimateInputs = [
+    "I'd love to select from the workshop tracks",
+    'We will create a table tennis tournament this year',
+    'Please update the event details into the new format',
+    'Drop into our open house anytime!',
+    'insert your team name into the field below',
+    'The theme color is 0xFF5733 for our banner',
+  ];
+
+  for (const input of legitimateInputs) {
+    await t.test(`allows: ${input}`, () => {
+      const { nextCalled, statusCode } = runGuard({ description: input });
+      assert.equal(nextCalled, true, 'next() should be called for legitimate input');
+      assert.equal(statusCode, null, 'no error status should be set');
+    });
+  }
+});
+
+test('sqlInjectionGuard still blocks real injection payloads', async (t) => {
+  const maliciousInputs = [
+    "1 OR 1=1",
+    "'; DROP TABLE users; --",
+    "SELECT * FROM users WHERE 1=1; WAITFOR DELAY '0:0:5'",
+    'UNION SELECT * FROM INFORMATION_SCHEMA.TABLES',
+  ];
+
+  for (const input of maliciousInputs) {
+    await t.test(`blocks: ${input}`, () => {
+      const { nextCalled, statusCode } = runGuard({ field: input });
+      assert.equal(nextCalled, false, 'next() should not be called for malicious input');
+      assert.equal(statusCode, 400, 'should respond with 400');
+    });
+  }
+});


Yuthika10:
Hi @Ayushh-Sharmaa! Can you review this PR and see if any further changes are needed to be made?

PR #3121, opened 2026-07-02
Fix Issue 2957: Missing Debounce on Directory/Event Search

### LINKED ISSUE #2957: [BUG]: Missing Debounce on Directory/Event Search leading to API Race Conditions and Unnecessary Re-renders.
### Bug Description

When typing into the search bar (whether searching for student portfolios or upcoming events), the application triggers a state update and network request for every single keystroke. On slower network connections, this causes a race condition where an older request might resolve after a newer one, displaying incorrect or outdated search results. It also causes noticeable UI lag due to rapid, unbatched React re-renders.

### Steps To Reproduce

Steps to Reproduce:
1.Navigate to the main Search/Explore page.
2.Throttle network speed to "Slow 3G" in browser DevTools.
3.Quickly type a query (e.g., "Web Development").
4.Watch the network tab: multiple requests are fired sequentially.
5.Observe the UI: the results may flicker or settle on results for a partial query (e.g., "Web Dev") rather than the full query, depending on which request returns last.

### Expected Behavior

Wrap the search input handler in a debounce function (e.g., using lodash.debounce or a custom React hook) with a delay of ~300-500ms so that the query only executes after the user pauses typing.

### Environment Info

Frontend (React Application)

Windows 11 

### Screenshots

Na

Goutam-2702:
# Summary
This PR fixes Issue #2957 by applying a debounce directly to the search input handler to avoid UI lag, and by solving a race condition in the search hook.

## Description
This PR addresses excessive re-renders and network race conditions caused by the directory and event search component. It wraps the input handler with a debounce function and adds a boolean abort guard for resolving pending API requests.

## Related Issue

Fixes #2957

## Type of Change

- [ ] Feature
- [x] Bug Fix
- [ ] UI/UX Improvement
- [ ] Security Enhancement
- [ ] Refactoring
- [ ] Documentation
- [ ] Testing
- [ ] Infrastructure
- [ ] Integration

## Changes Implemented
- Added a debounce to the search input handler in `SearchBar.jsx` to prevent excessive state updates and re-renders on every keystroke.
- Removed internal `useDebounce` from `useEventSearch.js` as it's now handled by the UI.
- Implemented an `active` abort flag in `useEventSearch.js` `useEffect` to guard against race conditions where older requests could resolve after newer requests, avoiding the display of incorrect or outdated search results.

### Frontend
- Used `useRef` and `setTimeout` in `SearchBar.jsx` to debounce the `onChange` event of the main search input.
- Added `active` flag within `useEffect` of `useEventSearch` hook to discard results if the effect is cleaned up due to subsequent queries.

### Unit Tests

- [ ]

### Integration Tests

- [ ]

### E2E Tests

- [ ]

### Manual Testing

- [x] Verified typing quickly does not cause immediate UI lag.
- [x] Verified search results display accurately for the final inputted query.
- [x] Simulated slow network and checked if the results correspond correctly to the query without older results overwriting newer ones.

## Performance Impact
Significant improvement to UI performance due to eliminated re-renders on every keystroke, reducing both React overhead and redundant network requests.

## Breaking Changes

- [x] No Breaking Changes

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Security reviewed
- [ ] Accessibility reviewed
- [x] Performance impact considered

### DIFF
--- a/website/src/components/SearchBar.jsx
+++ b/website/src/components/SearchBar.jsx
@@ -63,6 +63,22 @@ export default function SearchBar({ open, onClose, activities, events, onNavigat
     events
   );
 
+  const [localQuery, setLocalQuery] = useState('');
+  const timeoutRef = useRef(null);
+
+  const handleInputChange = (e) => {
+    const val = e.target.value;
+    setLocalQuery(val);
+    if (timeoutRef.current) clearTimeout(timeoutRef.current);
+    timeoutRef.current = setTimeout(() => {
+      setQuery(val);
+    }, 350);
+  };
+
+  useEffect(() => {
+    setLocalQuery(query);
+  }, [query]);
+
   useEffect(() => {
     if (open) {
       const t = setTimeout(() => inputRef.current?.focus(), 120);
@@ -169,8 +185,8 @@ export default function SearchBar({ open, onClose, activities, events, onNavigat
               <Search size={20} color="var(--c1)" style={{ flexShrink: 0 }} />
               <input
                 ref={inputRef}
-                value={query}
-                onChange={(e) => setQuery(e.target.value)}
+                value={localQuery}
+                onChange={handleInputChange}
                 placeholder="Search events, members, activities…"
                 aria-label="Search"
                 style={{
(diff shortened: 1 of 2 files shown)


Goutam-2702:
@Ayushh-Sharmaa pls review my pr and merge it.

PR #3658, opened 2026-07-13
fix: derive certificate IDs with HMAC instead of a plain hash

### LINKED ISSUE #3453: [BUG]: Certificate IDs are a deterministic, keyless hash allows enumeration and forgery of verification links
### Bug Description

In server-python/routers/certificates.py, the function make certificate id builds every certificate ID as sha256 of student_id and event_id joined by a colon, truncated to 16 hex characters, with no server side secret or salt involved.
Because GET /certificates/verify/certificate_id and GET /certificates/certificate_id/download are both public and unauthenticated, anyone who knows or can guess a student_id and event_id pair can compute the exact same certificate ID offline. Student IDs are often sequential or derivable from roll numbers or emails, and event IDs are public. This means an attacker can look up whether a specific student has a certificate for a specific event, download that certificate image which includes their full name and event name, and even predict a valid looking certificate ID for a student before any certificate has been issued for them.
This defeats the purpose of the verification flow. A certificate ID meant to prove authenticity should not be something a third party can compute without ever having received it.

### Steps To Reproduce

Pick any known student_id and event_id pair
Compute sha256 of student_id colon event_id, take first 16 hex characters, uppercase, prefix with NS-CERT-
Call GET /certificates/verify/that id
Observe that if a certificate exists for that pair, full student and event details are returned with no authentication required

### Expected Behavior

Certificate IDs should be unguessable without server-side secret knowledge. Verify and download should only return data to someone who was actually given the certificate ID, not to anyone who can guess a student_id/event_id pair. A certificate ID should also never resolve before the server has actually issued it.

### Environment Info

Python Forms Service (server-python)

Windows 11/Chrome 

### Screenshots

_No response_

vedikathalkar:
## Description
Certificate IDs were generated using a plain sha256 hash of student_id and event_id, with no server-side secret. Since the verify and download endpoints are public and unauthenticated, anyone who knew or guessed a student_id and event_id pair could compute the same certificate ID offline and look up or download that student's certificate without ever having been issued it.

## Related Issue
Fixes #3453

## Type of Change
- [x] Security Enhancement

## Changes Implemented
_make_certificate_id now derives the certificate ID using HMAC-SHA256 keyed with a new CERT_ID_SECRET environment variable, instead of an unkeyed hash. The ID stays deterministic, so the existing dedup logic in generate_certificates is unaffected, but it can no longer be computed without access to the server's secret. Also documented CERT_ID_SECRET in .env.example and README.md, following the same required-at-startup pattern already used for ADMIN_SECRET.

### Backend
Updated server-python/routers/certificates.py to use hmac.new keyed with CERT_ID_SECRET instead of hashlib.sha256 directly. Server now raises a RuntimeError on startup if CERT_ID_SECRET is not set, matching the existing ADMIN_SECRET check.

### Manual Testing
- [x] Verified the file parses correctly and the same student_id/event_id pair still produces a consistent certificate ID across calls, so the generate endpoint's dedup check is unaffected.

## Security Review
Certificate IDs can no longer be computed offline without knowledge of the server-side CERT_ID_SECRET.

## Breaking Changes
- [x] Breaking Changes Documented

Deployments must set CERT_ID_SECRET as a new required environment variable, or the server-python service will fail to start.

## Checklist
- [ ] Tests added or updated
- [x] Documentation updated
- [x] Security reviewed
- [ ] Accessibility reviewed
- [ ] Performance validated
- [ ] CI/CD passing
- [ ] Ready for review

### DIFF
--- a/server-python/.env.example
+++ b/server-python/.env.example
@@ -12,5 +12,9 @@ CORS_ORIGIN=http://localhost:5173,https://nexasphere-glbajaj.vercel.app
 # Admin secret for certificate generation API — set a strong random value
 ADMIN_SECRET=your-strong-random-secret
 
+# Secret used to derive certificate IDs (HMAC key) — set a strong random
+# value, and keep it different from ADMIN_SECRET
+CERT_ID_SECRET=your-strong-random-secret
+
 # Get your free Gemini API Key from: https://aistudio.google.com/app/apikey
 GEMINI_API_KEY=your_gemini_api_key_here
--- a/server-python/routers/certificates.py
+++ b/server-python/routers/certificates.py
@@ -45,6 +45,16 @@ if not ADMIN_SECRET:
         "Set it in your .env file or environment before starting the server."
     )
 
+# Secret used to derive certificate IDs. Must be kept server-side only —
+# anyone with this value could compute valid certificate IDs offline, so it
+# must never be the same value as ADMIN_SECRET or reused elsewhere.
+CERT_ID_SECRET = os.getenv("CERT_ID_SECRET")
+if not CERT_ID_SECRET:
+    raise RuntimeError(
+        "CERT_ID_SECRET environment variable is required to derive certificate IDs. "
+        "Set it in your .env file or environment before starting the server."
+    )
+
 
 def _require_admin(x_admin_secret: Optional[str] = Header(default=None)) -> None:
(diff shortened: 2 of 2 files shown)


vedikathalkar:
Hi @Ayushh-Sharmaa I have fixed via HMAC-SHA256 keyed with a new CERT_ID_SECRET instead of a plain hash.

PR #4085, opened 2026-07-29
fix: rewrite session middleware chain with idle timeout and activity tracking

### LINKED ISSUE #4080: [Bug] Session Middleware Chain Is Broken — Requests Hang Indefinitely
## Description
The session/logging middleware at lines 496-530 of `server/index.js` has its control flow completely corrupted by a bad merge. The `app.use((req, res, next) => { ... })` callback is declared at lines 496-512, but `next()` at line 511 is only called inside the `else if` branch (suspicious IP change), not in the normal branch (new session). In the normal path (lines 497-500), `next()` is **never called**, so requests hang until timeout. Lines 513-522 contain orphaned idle-timeout code that appears after the middleware function has been closed, causing `ReferenceError: req is not defined` at module load time. The tracking middleware at lines 524-530 never executes.

## Expected Behavior
All requests pass through session initialization, IP-change detection, idle timeout checks, and activity tracking.

## Actual Behavior
A `ReferenceError` crashes the server on load. If somehow avoided, the first middleware never calls `next()` in the normal path, causing all requests to hang indefinitely.

## Steps to Reproduce
Rewrite the entire middleware block (lines 496-530) as a single properly structured middleware that:
1. Always calls `next()` after session initialization (both new and existing sessions)
2. Moves the idle timeout check inside the middleware function body
3. Ensures proper session cleanup on timeout with a returned response
4. Restores the activity tracking middleware after the session middleware completes

## Affected Files
- `server/index.js` (lines 496-512) — Session middleware — `next()` call misplaced inside `else if` block
- `server/index.js` (lines 513-522) — Orphaned idle-timeout code that throws `ReferenceError`
- `server/index.js` (lines 524-530) — Tracking middleware that is unreachable

## Proposed Fix
Rewrite the entire middleware block (lines 496-530) as a single properly structured middleware that:
1. Always calls `next()` after session initialization (both new and existing sessions)
2. Moves the idle timeout check inside the middleware function body
3. Ensures proper session cleanup on timeout with a returned response
4. Restores the activity tracking middleware after the session middleware completes

ionfwsrijan:
## Fix: Restore broken session middleware chain

The session middleware never called `next()` in the normal path. Orphaned code after the middleware closure threw `ReferenceError`.

### Changes
- Created proper session middleware that always calls `next()` after initialization
- Moved idle timeout check inside middleware function body
- Restored activity tracking middleware after session middleware completes

Closes #4080

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -110,6 +110,8 @@ import * as studentAuthController from './controllers/studentAuthController.js';
 import * as forumController from './controllers/forumController.js';
 import { requireStudentAuth } from './middleware/studentAuthMiddleware.js';
 import { loadPersistedPushSubscriptions } from './routes/notifications.js';
+import { studentAuthService } from './services/studentAuthService.js';
+import { getAdminSession } from './repositories/adminSessionsRepository.js';
 import * as mentorshipController from './controllers/mentorshipController.js';
 import { xssSanitizer } from './middleware/xssSanitizer.js';
 import { tierRateLimiter } from './middleware/tierRateLimiter.js';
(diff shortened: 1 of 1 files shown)


ionfwsrijan:
@Ayushh-Sharmaa Please review this

PR #4087, opened 2026-07-29
fix: replace removeAllListeners with targeted socket.off calls, consolidate disconnect, add useAnalyticsSocket hook

### LINKED ISSUE #4082: [Bug] WebSocket Event Listener Accumulation and Socket Connection Leak Across Frontend
## Description
Multiple interconnected problems cause WebSocket resource leaks across the frontend:

**Problem A — Duplicate disconnect handlers & listener stomping** (`website/src/utils/socketClient.js`): Two `disconnect` event handlers are registered on the same event (lines 70-73 and 75-77). The cleanup code calls `removeAllListeners()` (line 47/49), destroying **all** listeners including those registered by other components — creating a stomping pattern where every re-initialization destroys other components' listeners.

**Problem B — Socket per mount** (`admin-dashboard/src/hooks/useAnalyticsSocket.js`): Uses `process.env.REACT_APP_SOCKET_URL` which is a CRA environment variable, but the admin dashboard uses **Vite** which requires `import.meta.env.VITE_*` prefix. Falls back to `'http://localhost:3001'` which is never the actual socket server. Creates a new `io()` connection every mount without reusing the existing singleton.

**Combined impact:** Every re-render destroys all listeners via `removeAllListeners()`. Every component mount creates a new WebSocket connection. This causes unbounded memory growth, file descriptor exhaustion, lost real-time events, and unnecessary network overhead.

## Expected Behavior
A single shared socket connection with properly scoped listener registration and cleanup.

## Actual Behavior
Each component creates its own socket connection; listener stomping causes lost events and memory leaks.

## Proposed Fix
1. **In `socketClient.js`**: Replace `removeAllListeners()` with targeted `socket.off()` calls that only remove the specific listeners registered by `initializeSocket`.
2. **In `useAnalyticsSocket.js`**: Rewrite to use the singleton from `socketClient.js` instead of creating new `io()` connections. Fix the Vite env var to `import.meta.env.VITE_SOCKET_URL`.

## Affected Files
- `website/src/utils/socketClient.js` (lines 46-53) — `removeAllListeners()` destroys other components' listeners
- `website/src/utils/socketClient.js` (lines 70-77) — Duplicate `disconnect` handlers
- `admin-dashboard/src/hooks/useAnalyticsSocket.js` (line 9) — Wrong env var prefix for Vite (always `undefined`)
- `admin-dashboard/src/hooks/useAnalyticsSocket.js` (lines 18-50) — Creates new `io()` connection on every mount
- `admin-dashboard/src/services/socketClient.js` — Provides `initAdminSocket()` singleton but `useAnalyticsSocket` ignores it

## Steps to Reproduce
1. **In `socketClient.js`**: Replace `removeAllListeners()` with targeted `socket.off()` calls that only remove the specific listeners registered by `initializeSocket`.
2. **In `useAnalyticsSocket.js`**: Rewrite to use the singleton from `socketClient.js` instead of creating new `io()` connections. Fix the Vite env var to `import.meta.env.VITE_SOCKET_URL`.

ionfwsrijan:
## Fix: Fix socket listener accumulation and connection leaks

`removeAllListeners()` destroyed other components' listeners. Components created new `io()` connections on every mount.

### Changes
- Replaced `removeAllListeners()` with targeted `socket.off()` for specific events only
- Consolidated duplicate disconnect handlers
- Fixed admin dashboard to use singleton from socketClient.js and corrected Vite env var prefix

Closes #4082

### DIFF
--- a/admin-dashboard/src/hooks/useAnalyticsSocket.js
+++ b/admin-dashboard/src/hooks/useAnalyticsSocket.js
@@ -1,163 +1,34 @@
-/**
- * Hook for Socket.IO analytics integration
- * Manages real-time connection and data updates
- */
+import { useState, useEffect, useCallback } from 'react';
+import { initAdminSocket, getSocket } from '../services/socketClient';
 
-import { useEffect, useState, useCallback, useRef } from 'react';
-import io from 'socket.io-client';
-
-const SOCKET_URL = process.env.REACT_APP_SOCKET_URL || 'http://localhost:3001';
-
-/**
- * Hook to manage analytics WebSocket connection
- */
-export function useAnalyticsSocket() {
-  const socketRef = useRef(null);
-  const [isConnected, setIsConnected] = useState(false);
+export function useAnalyticsSocket(eventId) {
+  const [analytics, setAnalytics] = useState(null);
+  const [connected, setConnected] = useState(false);
 
   useEffect(() => {
-    if (socketRef.current) return;
+    const socket = initAdminSocket();
+    if (!socket) return;
 
-    const socket = io(SOCKET_URL, {
-      transports: ['websocket'],
-      reconnection: true,
-      reconnectionDelay: 1000,
-      reconnectionDelayMax: 5000,
-      reconnectionAttempts: 5,
-    });
+    const onConnect = () => setConnected(true);
+    const onDisconnect = () => setConnected(false);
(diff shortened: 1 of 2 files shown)


ionfwsrijan:
@Ayushh-Sharmaa Please review this

PR #4672, opened 2026-08-12
fix(emailService): define attachments param and fix sendEmailOverride scope

### LINKED ISSUE #4629: [Bug] Undefined \ttachments\ variable in emailService.sendEmail causes every transactional email to fail silently
## Description

`server/services/emailService.js` `sendEmail()` references an `attachments` variable that is never declared, never imported, and never received as a parameter:

```js
export async function sendEmail({
  to,
  subject,
  templateName,
  data,
  from = defaultFrom,
  customTemplateContent = null,
}) {
  try {
    const html = await renderTemplate(templateName, data, customTemplateContent);
    const mailOptions = {
      from,
      to,
      subject,
      html,
      attachments,          // line 94 — ReferenceError: attachments is not defined
    };
    ...
    if (attachments.length > 0) {   // line 101 — same ReferenceError (dev branch)
      ...
    }
```

Evaluating `{ attachments }` throws `ReferenceError: attachments is not defined` **inside the `try`**, so the exception is swallowed by the catch (line 112) and every call resolves to `{ success: false }`.

Because `sendEmail` is the single choke point for all transactional mail:

- Welcome / email-verification emails (`sendWelcomeVerificationEmail`, called from `server/index.js:1009`)
- Password reset emails
- Waitlist / RSVP confirmations
- Event reminders and attendance emails
- Suspicious-login security alerts (`server/middleware/adminAuthMiddleware.js:596`)

all of these **silently fail** — the callers log a warning like `Failed to send welcome email` and continue, so there is no visible error but no email is ever delivered. Additionally, line 226 (`setSendEmailOverride`) assigns to an undeclared `sendEmailOverride` module variable.

## Expected Behavior

`sendEmail` accepts an optional `attachments` array, passes it to Nodemailer, and returns `{ success: true }` when the SMTP delivery succeeds. Transactional emails reach users.

## Actual Behavior

- Every `sendEmail` call throws `ReferenceError: attachments is not defined`, caught and converted to `{ success: false }`.
- All transactional emails are silently dropped across the platform; only error logs reveal the failure.
- The dev-mode "would send email" path also crashes on `attachments.length` before it can log.

## Proposed Fix

1. Accept `attachments` in the destructured parameters and default it to `[]`:

```js
export async function sendEmail({
  to,
  subject,
  templateName,
  data,
  from = defaultFrom,
  customTemplateContent = null,
  attachments = [],
}) {
```

2. Use it in `mailOptions` and guard the dev-mode logging:

```js
const mailOptions = { from, to, subject, html, attachments };
...
if (attachments.length > 0) {
  logger.info(`[Email Service - DEV] Attachments: ${attachments.map((a) => a.filename || 'attachment').join(', ')}`);
}
```

3. Fix `setSendEmailOverride` to use a module-level `let sendEmailOverride` (or pass the override through the transporter), so it does not assign to an undeclared variable.
4. Add a unit test that asserts `sendEmail` with a mocked transporter resolves `{ success: true }` and that `attachments` reach `sendMail`.

## Impact

High — the entire transactional email system is non-functional while reporting success; users never receive verification, reset, reminder, or security-alert emails.

## Files

- `server/services/emailService.js` (lines 78–120, 226)
- `server/index.js` (line 1009)
- `server/middleware/adminAuthMiddleware.js` (line 596)

ida-jemi:
## Description

Fixes a bug where `sendEmail()` in `server/services/emailService.js` referenced an undeclared `attachments` variable, throwing a `ReferenceError` on every call. Because the error was thrown inside the function's `try` block, it was silently caught and converted to `{ success: false }`, causing every transactional email on the platform (welcome/verification, password reset, RSVP/waitlist, event reminders, and security alerts) to fail silently.

Also fixes `setSendEmailOverride`, which was assigning to an undeclared module variable and was never actually consulted by `sendEmail`, breaking the SMTP-mocking hook used by `test/realtimeFaultIsolation.test.js`.

## Related Issue

Closes #4629

## Type of Change

- [x] Bug Fix
- [ ] Feature
- [ ] UI/UX Improvement
- [ ] Security Enhancement
- [ ] Refactoring
- [ ] Documentation
- [x] Testing
- [ ] Infrastructure
- [ ] Integration

## Changes Implemented

- Added `attachments = []` to `sendEmail`'s destructured parameters so it is properly defined and passed through to `mailOptions`.
- Declared `export let sendEmailOverride = null;` at module scope instead of relying on an implicit/undeclared global.
- Wired `sendEmail` to call `sendEmailOverride` when set, before hitting the real SMTP transporter, matching the existing pattern in `pushNotificationService.js`.
- Added `server/test/emailService.test.js` covering:
  - Attachments are correctly forwarded to the transporter.
  - Attachments default to `[]` without throwing.
  - `sendEmail` resolves `{ success: true }` on a mocked successful send.

### Backend

Changes are isolated to `server/services/emailService.js`:

- Updated the `sendEmail` function.
- Added the `sendEmailOverride` module variable.
- No API contract, route, or schema changes.

### Unit Tests

Added `server/test/emailService.test.js`.

- Verified the test fails with `ReferenceError: attachments is not defined` against the pre-fix code.
- Verified all tests pass after the fix.

### Manual Testing

Manually confirmed that `sendEmailOverride`, when set, is invoked and its thrown errors are caught and converted to `{ success: false }` as expected by `test/realtimeFaultIsolation.test.js`.

## Breaking Changes

- [x] No Breaking Changes

## Checklist

- [x] Tests added or updated
- [ ] Documentation updated
- [ ] Security reviewed
- [ ] Accessibility reviewed
- [ ] Performance validated
- [x] CI/CD passing
- [x] Ready for review

### DIFF
--- a/server/services/emailService.js
+++ b/server/services/emailService.js
@@ -32,6 +32,7 @@ const transporter = nodemailer.createTransport({
 });
 
 const defaultFrom = process.env.EMAIL_FROM || '"NexaSphere Team" <noreply@nexasphere.com>';
+export let sendEmailOverride = null;
 
 async function _sendMail(mailOptions) {
   return transporter.sendMail(mailOptions);
@@ -82,6 +83,7 @@ export async function sendEmail({
   data,
   from = defaultFrom,
   customTemplateContent = null,
+  attachments = [],
 }) {
   try {
     const html = await renderTemplate(templateName, data, customTemplateContent);
@@ -93,6 +95,12 @@ export async function sendEmail({
       html,
       attachments,
     };
+
+    if (sendEmailOverride) {
+      const info = await sendEmailOverride(mailOptions);
+      logger.info(`[Email Service] Message sent successfully via override: ${info?.messageId || 'n/a'}`);
+      return { success: true, messageId: info?.messageId };
+    }
 
     if (!isProduction && (!process.env.SMTP_USER || !process.env.SMTP_PASS)) {
       logger.info(`[Email Service - DEV] Would send email to: ${to}`);
--- a/server/test/emailService.test.js
+++ b/server/test/emailService.test.js
@@ -1,0 +1,57 @@
+import assert from 'node:assert/strict';
+import test, { mock } from 'node:test';
+
+// Exercise the real SMTP path (not the dev-mode simulation branch) via a
+// mocked nodemailer transporter instead of a live SMTP connection.
+process.env.SMTP_HOST = 'smtp.test.local';
+process.env.SMTP_PORT = '587';
+process.env.SMTP_USER = 'test-user';
+process.env.SMTP_PASS = 'test-pass';
+
+const sendMailCalls = [];
+
+mock.module('nodemailer', {
+  defaultExport: {
+    createTransport: () => ({
+      sendMail: async (mailOptions) => {
+        sendMailCalls.push(mailOptions);
+        return { messageId: 'mock-message-id' };
+      },
+    }),
+  },
+});
+
+const { sendEmail } = await import('../services/emailService.js');
+
+test('sendEmail resolves { success: true } and forwards attachments to the transporter', async () => {
+  sendMailCalls.length = 0;
+
+  const attachments = [{ filename: 'invoice.pdf', content: 'dummy-content' }];
+
+  const result = await sendEmail({
+    to: 'user@example.com',
+    subject: 'Welcome to NexaSphere',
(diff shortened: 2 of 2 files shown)


ida-jemi:
Hi @Ayushh-Sharmaa,
I’ve contributed to this project as part of GSSoC’26. Could you please add the relevant GSSoC’26 and other necessary labels to this PR? Thank you!

